PR #4, opened 2020-10-07
Port autoware_localization_srvs

nnmm:
  * running `colcon build --packages-up-to autoware_localization_srvs` in the `ros2` branch of the AutowareArchitectureProposal repo (with this branch checked out in the `autoware/autoware.iv` sub-repo)

### DIFF
--- a/localization/util/autoware_localization_srvs/CMakeLists.txt
+++ b/localization/util/autoware_localization_srvs/CMakeLists.txt
@@ -15,6 +15,7 @@ find_package(geometry_msgs REQUIRED)
 
 rosidl_generate_interfaces(${PROJECT_NAME}
   "srv/PoseWithCovarianceStamped.srv"
+  DEPENDENCIES geometry_msgs
 )
 
 ament_package()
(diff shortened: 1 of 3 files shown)


nnmm:
@TakaHoribe please review!
The only interesting things about this were:
* `CMakeLists.txt` didn't mention `geometry_msgs` even though the package depends on it in `package.xml`. I added it there, but removed `std_msgs` from `package.xml` since transitive dependencies don't need to specified.
* Also, I'm wondering why idl packages have `<member_of_group>rosidl_interface_packages</member_of_group>` – is there an implicit `<group_depend>rosidl_interface_packages</group_depend>` somewhere?

TakaHoribe:
@nnmm Thank you, Nicolai! This can be built in my environment.

 - It seems the `geometry_msgs` is already mentioned in the original code. The `std_msgs` can be removed.

 - I don't know much about the `<member_of_group>rosidl_interface_packages</member_of_group>`. It is required for the minimum configuration of ros2 message generation.

TakaHoribe (approved):
Confirmed it worked in my env correctly.

PR #7, opened 2020-10-07
Port shift decider to ros2

fred-apex-ai:
I ported the code similar in a similar fashion to what @nnmm did with the `vehicle_cmd_gate` and took the review comments given on that [PR](https://github.com/tier4/autoware.iv.universe/pull/3). The PR currently suffers from the same limitation that a jump in sim time would be treated differently than in the previous version. Before implementing a hack, we should agree on a common solution and then implement it in all packages to be ported.

The first commit of this PR is the previous ROS1 version of the code.
 
The node builds and runs fine with 

    ros2 launch shift_decider shift_decider.launch.xml

@nnmm @dejanpan @mitsudome-r

### DIFF
--- a/control/shift_decider/src/shift_decider.cpp
+++ b/control/shift_decider/src/shift_decider.cpp
@@ -16,33 +16,42 @@
 
 #include <shift_decider/shift_decider.h>
 
-ShiftDecider::ShiftDecider()
+#include <cstddef>
+#include <functional>
+
+ShiftDecider::ShiftDecider() : Node("shift_decider")
 {
-  pub_shift_cmd_ = pnh_.advertise<autoware_vehicle_msgs::ShiftStamped>("output/shift_cmd", 1, true);
-  sub_control_cmd_ = pnh_.subscribe("input/control_cmd", 1, &ShiftDecider::onControlCmd, this);
-  timer_ = nh_.createTimer(ros::Duration(0.1), &ShiftDecider::onTimer, this);
+  using std::placeholders::_1;
+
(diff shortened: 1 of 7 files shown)


control/shift_decider/src/shift_decider.cpp
```diff
@@ -16,33 +16,42 @@
...
-  timer_ = nh_.createTimer(ros::Duration(0.1), &ShiftDecider::onTimer, this);
+  using std::placeholders::_1;
+
+  static constexpr std::size_t queue_size = 1;
+  rclcpp::QoS durable_qos(queue_size);
+  durable_qos.transient_local();
+
+  pub_shift_cmd_ = create_publisher<autoware_vehicle_msgs::msg::ShiftStamped>("output/shift_cmd", durable_qos);
+  sub_control_cmd_ = create_subscription<autoware_control_msgs::msg::ControlCommandStamped>("input/control_cmd", queue_size, std::bind(&ShiftDecider::onControlCmd, this, _1));
+  timer_ = create_wall_timer(std::chrono::duration<double>(0.1), std::bind(&ShiftDecider::onTimer, this));
```
mitsudome-r, line 32: We have the same issue with vehicle_cmd_gate regarding timer.
Once investigation is done, we can fix this if needed.

fred-apex-ai: @mitsudome-r I added the timer solution as in the `vehicle_cmd_gate`. Please have a final look, hopefully we can then merge this PR

mitsudome-r (approved):
LGTM

PR #11, opened 2020-10-08
Port pose_initializer to ROS2

nnmm:
  * running `colcon build --packages-up-to pose_initializer` in the `ros2` branch of the AutowareArchitectureProposal repo (with this branch checked out in the `autoware/autoware.iv` sub-repo)
  * running the launch file

### DIFF
--- a/localization/util/pose_initializer/CMakeLists.txt
+++ b/localization/util/pose_initializer/CMakeLists.txt
@@ -19,7 +19,7 @@ find_package(tf2_geometry_msgs REQUIRED)
 find_package(std_msgs REQUIRED)
 find_package(geometry_msgs REQUIRED)
 find_package(sensor_msgs REQUIRED)
-find_package(pcl_ros REQUIRED)
+find_package(PCL REQUIRED)
 find_package(pcl_conversions REQUIRED)
 find_package(autoware_localization_srvs REQUIRED)
 
@@ -39,7 +39,7 @@ set(POSE_INITIALIZER_DEPENDENCIES
   std_msgs
   geometry_msgs
   sensor_msgs
-  pcl_ros
+  PCL
   pcl_conversions
   autoware_localization_srvs
 )
--- a/localization/util/pose_initializer/CMakeLists.txt
+++ b/localization/util/pose_initializer/CMakeLists.txt
@@ -57,3 +57,5 @@ install(
     launch
   DESTINATION share/${PROJECT_NAME}
 )
+
+ament_package()
--- a/localization/util/pose_initializer/CMakeLists.txt
+++ b/localization/util/pose_initializer/CMakeLists.txt
@@ -1,47 +1,59 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(pose_initializer)
 
-add_compile_options(-std=c++14)
+### Compile options
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wno-unused-parameter)
+endif()
 
-find_package(catkin REQUIRED COMPONENTS
-  roscpp
-  tf2
-  tf2_ros
-  tf2_geometry_msgs
-  std_msgs
-  geometry_msgs
-  sensor_msgs
-  dynamic_reconfigure
-  pcl_ros
-  pcl_conversions
-  autoware_localization_srvs
-)
+### Dependencies
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(geometry_msgs REQUIRED)
+find_package(sensor_msgs REQUIRED)
+find_package(pcl_ros REQUIRED)
+find_package(pcl_conversions REQUIRED)
+find_package(autoware_localization_srvs REQUIRED)
 
-catkin_package(
-  INCLUDE_DIRS include
-)
-
-include_directories(include ${catkin_INCLUDE_DIRS})
+### Includes
+include_directories(include)
 
 add_executable(pose_initializer
   src/pose_initializer_node.cpp
   src/pose_initializer_core.cpp
 )
 
-add_dependencies(pose_initializer ${${PROJECT_NAME}_EXPORTED_TARGETS} ${catkin_EXPORTED_TARGETS})
+set(POSE_INITIALIZER_DEPENDENCIES
+  rclcpp
+  tf2
+  tf2_ros
+  tf2_geometry_msgs
+  std_msgs
+  geometry_msgs
+  sensor_msgs
+  pcl_ros
+  pcl_conversions
+  autoware_localization_srvs
+)
 
-target_link_libraries(pose_initializer ${catkin_LIBRARIES})
+ament_target_dependencies(pose_initializer ${POSE_INITIALIZER_DEPENDENCIES})
+ament_export_dependencies(${POSE_INITIALIZER_DEPENDENCIES})
 
 install(
-  TARGETS
-    pose_initializer
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+  TARGETS pose_initializer
+  DESTINATION lib/${PROJECT_NAME}
 )
 
 install(
   DIRECTORY
     launch
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+  DESTINATION share/${PROJECT_NAME}
 )
--- a/localization/util/pose_initializer/src/pose_initializer_core.cpp
+++ b/localization/util/pose_initializer/src/pose_initializer_core.cpp
@@ -179,9 +179,13 @@ bool PoseInitializer::getHeight(
 void PoseInitializer::callAlignServiceAndPublishResult(
   const geometry_msgs::msg::PoseWithCovarianceStamped::SharedPtr input_pose_msg)
 {
+  if (request_id_ != response_id_) {
+    throw std::runtime_error("Did not receive response for previous NDT Align Server call");
+  }
   auto req =
     std::make_shared<autoware_localization_srvs::srv::PoseWithCovarianceStamped::Request>();
   req->pose_with_cov = *input_pose_msg;
+  req->seq = ++request_id_;
 
   RCLCPP_INFO(get_logger(), "call NDT Align Server");
 
@@ -190,6 +194,7 @@ void PoseInitializer::callAlignServiceAndPublishResult(
     [this](rclcpp::Client<autoware_localization_srvs::srv::PoseWithCovarianceStamped>::SharedFuture
              result) {
       RCLCPP_INFO(get_logger(), "called NDT Align Server");
+      response_id_ = result.get()->seq;
       // NOTE temporary cov
       geometry_msgs::msg::PoseWithCovarianceStamped & pose_with_cov = result.get()->pose_with_cov;
       pose_with_cov.pose.covariance[0] = 1.0;
--- a/localization/util/pose_initializer/src/pose_initializer_core.cpp
+++ b/localization/util/pose_initializer/src/pose_initializer_core.cpp
@@ -64,7 +64,9 @@ PoseInitializer::PoseInitializer()
 
   ndt_client_ = this->create_client<autoware_localization_srvs::srv::PoseWithCovarianceStamped>(
     "ndt_align_srv");
-  ndt_client_->wait_for_service(std::chrono::seconds(1));  // TODO
+  while (!ndt_client_->wait_for_service(std::chrono::seconds(1)) && rclcpp::ok()) {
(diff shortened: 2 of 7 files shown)


localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -167,11 +172,10 @@ bool PoseInitializer::getHeight(
   if (map_ptr_) {
     tf2::Transform transform;
     try {
-      const auto stamped =
-        tf2_buffer_.lookupTransform(map_frame_, fixed_frame, ros::Time(0), ros::Duration(1.0));
+      const auto stamped = tf2_buffer_.lookupTransform(map_frame_, fixed_frame, tf2::TimePointZero);
```
nnmm, line 175: There is a slight semantic difference here – the timeout doesn't matters if the tf buffer contains at least one pose, no matter the timestamp. If this timeout is here intentionally for the case where the tf buffer might not contain any pose, please let me know.

localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -146,18 +152,17 @@ void PoseInitializer::callbackGNSSPoseCov(
   add_height_pose_msg_ptr->pose.covariance[4 * 6 + 4] = 0.01;
   add_height_pose_msg_ptr->pose.covariance[5 * 6 + 5] = 3.14;
 
-  geometry_msgs::PoseWithCovarianceStamped::Ptr aligned_pose_msg_ptr(
-    new geometry_msgs::PoseWithCovarianceStamped);
-  const bool succeeded_align = callAlignService(*add_height_pose_msg_ptr, aligned_pose_msg_ptr);
+  auto aligned_pose_msg_ptr = std::make_shared<geometry_msgs::msg::PoseWithCovarianceStamped>();
```
nnmm, line 155: There is a possible improvement here that I didn't implement to keep it focused on the ROS2 port – this could live on the stack instead, and then be used as an out parameter by passing in a raw pointer.

localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -188,26 +192,33 @@ bool PoseInitializer::getHeight(
...
+    pose_with_cov.pose.covariance[1 * 6 + 1] = 1.0;
+    pose_with_cov.pose.covariance[2 * 6 + 2] = 0.01;
+    pose_with_cov.pose.covariance[3 * 6 + 3] = 0.01;
+    pose_with_cov.pose.covariance[4 * 6 + 4] = 0.01;
+    pose_with_cov.pose.covariance[5 * 6 + 5] = 0.2;
+    *output_pose_msg_ptr = pose_with_cov;
     return true;
   } else {
-    ROS_ERROR("[pose_initializer] could not call NDT Align Server");
+    RCLCPP_ERROR(get_logger(), "could not call NDT Align Server");
```
nnmm, line 221: The logging library should already prepend "[pose_initializer]" to the message

localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -88,25 +99,22 @@ bool PoseInitializer::serviceInitial(
...
+  auto aligned_pose_msg_ptr = std::make_shared<geometry_msgs::msg::PoseWithCovarianceStamped>();
+  const bool succeeded_align = callAlignService(add_height_pose_msg_ptr, aligned_pose_msg_ptr);
 
   if (succeeded_align) {
-    initial_pose_pub_.publish(*aligned_pose_msg_ptr);
-    return true;
+    initial_pose_pub_->publish(*aligned_pose_msg_ptr);
   } else {
-    return false;
+    throw std::runtime_error("Aligning poses did not succeed.");
```
nnmm, line 108: Service handlers don't return a bool anymore. It is recommended by the migration guide to throw an exception instead.

localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -38,47 +41,55 @@ double getGroundHeight(const pcl::PointCloud<pcl::PointXYZ>::Ptr pcdmap, const t
...
-  initial_pose_pub_ = nh_.advertise<geometry_msgs::PoseWithCovarianceStamped>("initialpose3d", 10);
+  initial_pose_pub_ =
+    this->create_publisher<geometry_msgs::msg::PoseWithCovarianceStamped>("initialpose3d", 10);
 
-  ndt_client_ =
-    nh_.serviceClient<autoware_localization_srvs::PoseWithCovarianceStamped>("ndt_align_srv");
-  ndt_client_.waitForExistence(ros::Duration(1.0));  // TODO
+  ndt_client_ = this->create_client<autoware_localization_srvs::srv::PoseWithCovarianceStamped>(
+    "ndt_align_srv");
+  ndt_client_->wait_for_service(std::chrono::seconds(1));  // TODO
```
nnmm, line 67: Not sure if the TODO refers to that, but this function returns a bool that could be used to check if the service went up.

localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -188,26 +192,33 @@ bool PoseInitializer::getHeight(
...
+  auto req =
+    std::make_shared<autoware_localization_srvs::srv::PoseWithCovarianceStamped::Request>();
+  req->pose_with_cov = *input_pose_msg;
+
+  RCLCPP_INFO(get_logger(), "call NDT Align Server");
+
+  auto result = ndt_client_->async_send_request(req);
+  // Wait for the result. `shared_from_this` is fine since this class is only used with `make_shared`.
+  if (
+    rclcpp::spin_until_future_complete(this->shared_from_this(), result) ==
```
mitsudome-r, line 207: We had a problem with Autoware.Auto where we couldn't do spin_until_future_complete() inside a Node class.
The node itself is already added to the executor in spin() function inside the main function, and spin_until_future_complete also tries to add the node to executor inside the function. Since executor throws error when the user tries to add same node twice, it will abort when process reach to this line.

mitsudome-r: Workaround is to call async_send_request with a callback, or handover the executor to the PoseInitializer and do the similar thing that is done in spin_until_future_complete() function by ourselves.

mitsudome-r: We might want to discuss about this in the meeting on Monday.

mitsudome-r (commented):
I think the porting is done well overall, but we have a problem with service call where we don't have complete replacement for synchronous call in ROS2.

mitsudome-r:
@nnmm Also, we have updated CI so could you try rebasing your branch to the newest ros2 branch? Thanks.

localization/util/pose_initializer/CMakeLists.txt
```diff
@@ -1,47 +1,61 @@
...
+
+set(POSE_INITIALIZER_DEPENDENCIES
+  rclcpp
   tf2
   tf2_ros
   tf2_geometry_msgs
   std_msgs
   geometry_msgs
   sensor_msgs
-  dynamic_reconfigure
```
nnmm, line 14: dynamic_reconfigure shouldn't be needed anymore in ROS2

localization/util/pose_initializer/CMakeLists.txt
```diff
@@ -1,47 +1,61 @@
...
-  pcl_ros
```
nnmm, line 15: None of the methods provided by pcl_ros were actually used, it could be replaced with PCL until the port of pcl_ros is done.

localization/util/pose_initializer/src/pose_initializer_core.cpp
nnmm: Thanks for pointing that out! How about just calling `wait()` or `get()` on the returned future? I can't imagine that interfering with the node, or the design would be broken.

nnmm: Okay, those just block forever. I rewrote it with a callback, which was fortunately easy. The drawback is that there is no way to detect failure there: If the service is offline, the callback is just never called. There's no good way to set a timeout either.

I wonder if `.wait()` would work with a `MultiThreadedExecutor`.

nnmm:
@mitsudome-r This should now be ready for final review – I'm mentioning you here since you already had a look, but feel free to reassign.

I commented a few interesting bits. The main questions are whether https://github.com/tier4/Pilot.Auto/pull/11#discussion_r501788711 and https://github.com/tier4/Pilot.Auto/pull/11#discussion_r503254078 are acceptable to you.

localization/util/pose_initializer/src/pose_initializer_core.cpp
mitsudome-r: I think this is okay if we are using Time(0) since it is looking for the newest transform.
However, I think we have some nodes that require lookupTransform with exact time, which will need some timeout.
If you have any experience in using tf2::BufferClient (which seems to have lookupTransform with timeout), could you comment on this issue?https://github.com/tier4/Pilot.Auto/issues/23

localization/util/pose_initializer/CMakeLists.txt
mitsudome-r: :+1:

localization/util/pose_initializer/CMakeLists.txt
mitsudome-r: Since we decided to remove unused packages, this is okay.
https://github.com/tier4/AutowareArchitectureProposal/issues

localization/util/pose_initializer/src/pose_initializer_core.cpp
mitsudome-r: Yes, I think this TODO refers to error check.
Since we have no error handling when we call the service anymore, we might want to change this so that it waits until the service is actually available.

e.g.
```cpp
  while (!ndt_client_->wait_for_service(std::chrono::seconds(1)) && rclcpp::ok()) {
    RCLCPP_INFO(get_logger(), "Waiting for service...");
  }
```

localization/util/pose_initializer/src/pose_initializer_core.cpp
mitsudome-r: It might work with MultiThreadedExecutor, but I think it would cause larger impact since we would then have to think about using mutex::lock in callbacks.

mitsudome-r:
@nnmm 
I think we can merge this after this is resolved https://github.com/tier4/Pilot.Auto/pull/11#discussion_r505481898.

CI should pass after https://github.com/tier4/Pilot.Auto/pull/31 is merged.
Also, if you want to rerun CI without adding empty commits, you can go to "Checks" tab and click "re-run jobs" at right top.

image

localization/util/pose_initializer/src/pose_initializer_core.cpp
nnmm: If you prefer, we could at least have a minimal check like this:
* have a member variable that identifies the latest response received and one for the latest request sent
* set the response identifier in the callback
* before the next client call, check that the response identifier corresponds to the request identifier, and then set the request variable

localization/util/pose_initializer/src/pose_initializer_core.cpp
nnmm: :+1: Done.

localization/util/pose_initializer/src/pose_initializer_core.cpp
nnmm: I think even with Time(0) there is a corner case where a timeout is relevant: If an initial pose message or service call occurs slightly before whatever node is responsible has published the first transform, it will fail, i.e. throw an exception.

localization/util/pose_initializer/src/pose_initializer_core.cpp
mitsudome-r: I see. that sounds good to me.
Since it is now a callback and does not going to block the process, we probably have to make sure that we don't send new service request before previous request is done.

mitsudome-r:
@nnmm do you want to work on this? If not, that's also okay, and I can merge this PR.
https://github.com/tier4/Pilot.Auto/pull/11#discussion_r505525278

nnmm:
@mitsudome-r No, I'll see about implementing that later. Sorry this took a bit longer.

localization/util/pose_initializer/src/pose_initializer_core.cpp
nnmm: Done. I added a sequence number to the request, and in the ndt_scan_matcher I will need to copy that number from the request into the response.

localization/util/pose_initializer/src/pose_initializer_core.cpp
```diff
@@ -187,27 +176,33 @@ bool PoseInitializer::getHeight(
...
-    srv.response.pose_with_cov.pose.covariance[3 * 6 + 3] = 0.01;
-    srv.response.pose_with_cov.pose.covariance[4 * 6 + 4] = 0.01;
-    srv.response.pose_with_cov.pose.covariance[5 * 6 + 5] = 0.2;
-    *output_pose_msg_ptr = srv.response.pose_with_cov;
-    return true;
-  } else {
-    ROS_ERROR("[pose_initializer] could not call NDT Align Server");
-    return false;
+  if (request_id_ != response_id_) {
+    throw std::runtime_error("Did not receive response for previous NDT Align Server call");
```
mitsudome-r, line 183: Throwing an error and terminating the program might be too conservative.
We can modify this so that it only outputs RCLCPP_ERROR message and returns.
That way, users can just try initializing again after waiting for service call to finish without restarting the node.

mitsudome-r:
```suggestion
    RCLCPP_ERROR(get_logger(), "Did not receive response for previous NDT Align Server call");
    return;
```

nnmm: Sure, done.

localization/util/pose_initializer/src/pose_initializer_core.cpp
mitsudome-r: Thanks. Looks good to me.

mitsudome-r (approved):

nnmm:
Thanks @mitsudome-r!

PR #18, opened 2020-10-11
ROS2 Porting: gnss_poser

jilaada:
# gnss_poser

Awaiting on the following PRs:
* [x] https://github.com/tier4/AutowareArchitectureProposal/pull/69

## Summary

Related PRs:
* https://github.com/tier4/ublox/pull/2
* https://github.com/tier4/AutowareArchitectureProposal/pull/69

Convert the gnss_poser package to ROS2. Requires a number of dependencies which are not in the docker:
* ublox_msgs
* geographic_lib
* geo_pos_conv (already ported)

## Testing

Testing this branch will require some dependencies which are in preinstalled in Foxy:
* GeographicLib (installed via rosdep)
* diagnostic_updater (installed via rosdep)
* ublox_msgs (build and install)

Setup:
1. Checkout PR branch for AutowareArchitectureProposal repository (https://github.com/tier4/AutowareArchitectureProposal/pull/69)
2. Modify the reference to the ublox to point to the PR branch (https://github.com/tier4/ublox/pull/2) - since this has not already been merged
3. Run the vcs tool
```
vcs import src < autoware.proj.repos
```
4. Enter your environment of choice (docker/ADE/etc.)
5. Install rosdep dependencies
```
sudo apt update
rosdep install -y --from-paths /path/to/AutowareArchitectureProposal/src/ --ignore-src --rosdistro foxy
```
6. Build and source the repository (the build will `stderr` see Discussion Points below)
```
colcon build
. install/setup.bash
```
7. Run the launch files and check the topics, parameters and node exists (two launch files total)
```
ros2 launch gnss_poser gnss_poser.launch.xml
ros2 launch gnss_poser ubloxfix2mgrs.launch.xml
```

### Logging
Non-ROS-node helper method originally had some ROS logging method calls. To convert this to ROS2 while maintaining the ROS-based logging I needed to pass the node's logger by reference to the methods. Not sure if this is the tried and vetted method of maintaining logging in non-ROS helper methods. Please advise.

### Ublox dependencies
The Ublox dependencies are stored in the `/sensing/driver/ublox` but in the AutowareArchitectureProposal repo the `.repos` file maps the ublox repository to a folder much higher. I noticed that the ublox repository is neither a submodule or pulled via the `.repos` is there a way we want to try and pull this resource?


jilaada:
I didn't start with using components or using ament_auto for building packages - since discussions have been in favour of these I shall make those changes here.

jilaada:
I have made the following updates to this PR:
* Convert ament_cmake to ament_cmake_auto resulting in clean up in the CMakelist
* Convert to use components
  * Remove gnss_poser_core -> gnss_poser_node (removing the main function previously existing in gnss_poser_node) - sorry in advance I think this has made it hard to review due to the renaming but I found the naming otherwise confusing if this gets too difficult to review I will happily revert the file name changes.
* Expand header guards to use the ROS standard
* Add -Werror compile flag

The build still fails due to not being able to find the ublox dependencies, these will need to be compiled on your system manually using the tier4 ublox repository. There has been a release yesterday with a foxy version of the ublox package, however it seems that rosdep is still not able to find it yet

nnmm:
I didn't get around to reviewing yesterday, but I'm taking a look now

jilaada:
@nnmm Thanks for reviewing I just patched it to allow the node to be launched and reverted the name changes so that the review process would be easier, sorry for all the back and forth.

sensing/preprocessor/gnss/gnss_poser/package.xml
```diff
@@ -1,87 +1,35 @@
...
+  <depend>sensor_msgs</depend>
+  <depend>geo_pos_conv</depend>
+  <depend>ublox_msgs</depend>
+
+  <exec_depend>rclcpp</exec_depend>
+  <exec_depend>rclcpp_components</exec_depend>
+  <exec_depend>launch_ros</exec_depend>
+  <exec_depend>rosidl_default_runtime</exec_depend>
   <exec_depend>geographiclib-tools</exec_depend>
+  <exec_depend>geographiclib</exec_depend>
```
nnmm, line 30: Isn't this also a build-time dependency, i.e. `<depend>`?

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -109,18 +91,20 @@ void GNSSPoser::callbackNavSatFix(const sensor_msgs::NavSatFix::ConstPtr & nav_s
...
+  geometry_msgs::msg::PoseStamped gnss_antenna_pose_msg;
   gnss_antenna_pose_msg.header.stamp = nav_sat_fix_msg_ptr->header.stamp;
   gnss_antenna_pose_msg.header.frame_id = map_frame_;
   gnss_antenna_pose_msg.pose.position = median_position;
   gnss_antenna_pose_msg.pose.orientation = orientation;
 
   // get TF from base_link to gnss_antenna
-  geometry_msgs::TransformStamped::Ptr TF_base_to_gnss_ptr(new geometry_msgs::TransformStamped);
-  getStaticTransform(gnss_frame_, base_frame_, TF_base_to_gnss_ptr, nav_sat_fix_msg_ptr->header.stamp);
+  geometry_msgs::msg::TransformStamped::SharedPtr TF_base_to_gnss_ptr(
```
nnmm, line 101: Nitpick: If you are using the Google Style guide, variable names should not be uppercase. Also, make_shared can be used here.

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -13,76 +13,57 @@
...
   tf2_listener_(tf2_buffer_),
-  coordinate_system_(CoordinateSystem::MGRS),
-  base_frame_("base_link"),
-  gnss_frame_("gnss"),
-  gnss_base_frame_("gnss_base_link"),
-  map_frame_("map"),
-  use_ublox_receiver_(false),
-  plane_zone_(9)
+  tf2_broadcaster_(this),
+  clock_(RCL_ROS_TIME)
```
nnmm, line 26: I see that this is only used to timestamp log messages, but does this work? I think you might have to use `Node::get_clock()` because the clock needs to be told by the node when sim time is enabled. I can test this later.

nnmm (approved):
Looks good!

sensing/preprocessor/gnss/gnss_poser/CMakeLists.txt
```diff
@@ -1,66 +1,48 @@
...
-)
+### Compile options
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wno-unused-parameter -Wall -Wextra -Wpedantic -Werror)
+endif()
+
+find_package(ament_cmake_auto)
```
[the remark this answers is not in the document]
jilaada: I can try test this, though my concern is with some of the external packages it pulls in such as geographiclib. I'll remove the redundant calls if I am successful!

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -13,76 +13,57 @@
...
+    declare_parameter("coordinate_system", static_cast<int>(CoordinateSystem::MGRS));
   coordinate_system_ = static_cast<CoordinateSystem>(coordinate_system);
 
-  private_nh_.getParam("base_frame", base_frame_);
-  private_nh_.getParam("gnss_frame", gnss_frame_);
-  private_nh_.getParam("gnss_base_frame", gnss_base_frame_);
-  private_nh_.getParam("map_frame", map_frame_);
-  private_nh_.getParam("use_ublox_receiver", use_ublox_receiver_);
-  private_nh_.getParam("plane_zone", plane_zone_);
+  base_frame_ = declare_parameter("base_frame", "base_link");
```
fred-apex-ai, line 32: I looked at `declare_parameter` yesterday myself in porting `mpc_follower`. Haven't gotten the whole code to compile so I could test this myself but it looks ok. If you want to support only the use-case to set values initially through e.g. the launch file, then, however, we should  declare the members like `gnss_frame` as `const` and move the initialization to the initializer list:

```c++
GNSSPoser::GNSSPoser(const rclcpp::NodeOptions & node_options) :
...
gnss_frame(declare_parameter("gnss_frame", "gnss"))
```

But again, this is a suggestion that goes slightly beyond porting alone

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
fred-apex-ai: good catch. Since we use c++14, we should use `make_shared` for exception safety and to save potentially one memory allocation Good post [here](https://arne-mertz.de/2018/09/make_shared-vs-the-normal-shared_ptr-constructor/)

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -265,13 +250,14 @@ bool GNSSPoser::getTransform(
...
-    ROS_WARN_STREAM_THROTTLE(1, ex.what());
-    ROS_WARN_STREAM_THROTTLE(
-      1, "Please publish TF " << target_frame.c_str() << " to " << source_frame.c_str());
+    RCLCPP_WARN_STREAM_THROTTLE(this->get_logger(), clock_, 1, ex.what());
+    RCLCPP_WARN_STREAM_THROTTLE(
+      this->get_logger(), clock_, 1,
+      "Please publish TF " << target_frame.c_str() << " to " << source_frame.c_str());
 
-    transform_stamped_ptr->header.stamp = ros::Time::now();
+    transform_stamped_ptr->header.stamp = rclcpp::Node::now();
```
fred-apex-ai, line 260: One of those methods that are [not documented](https://docs.ros2.org/dashing/api/rclcpp/classrclcpp_1_1Node.html#af706b231620f1c120b7ccd738ec31867) 
I'm not sure which time it returns, I"m guessing it's the system time as it seems to be a static method. If you respect @nnmm 's comment above, then I think you should ask the `clock_` member for `now()` to serve the sim and replay use cases as well

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -351,4 +339,4 @@ void GNSSPoser::publishTF(
   tf2_broadcaster_.sendTransform(transform_stamped);
 }
 
-}  // namespace GNSSPoser
+RCLCPP_COMPONENTS_REGISTER_NODE(GNSSPoser)
```
fred-apex-ai, line 342: I like this pattern to get of the hand-written main function!

fred-apex-ai (changes requested):

sensing/preprocessor/gnss/gnss_poser/launch/ubloxfix2mgrs.launch.xml
```diff
@@ -1,12 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <launch>
-
   <arg name="input_topic_fix" default="/ublox/fix" />
   <arg name="input_topic_navpvt" default="/ublox/navpvt" />
 
-  <include file="$(find gnss_poser)/launch/gnss_poser.launch" >
-    <arg name="input_topic_fix" value="$(arg input_topic_fix)" />
-    <arg name="input_topic_navpvt" value="$(arg input_topic_navpvt)" />
+  <include file="$(find-pkg-share gnss_poser)/launch/gnss_poser.launch.xml" >
```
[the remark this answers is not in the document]
jilaada: i haven't looked too deeply into it but I used this [reference doc](https://index.ros.org/doc/ros2/Tutorials/Launch-files-migration-guide/#substitutions) and it seems that the keyword `find` is replaced with `find-pkg-share` or `find-pkg-prefix` (shared directory vs root installed package, respectively).

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
jilaada: These are good suggestions I will add them, though they are going slightly beyond a porting. I think the parameters are unlikely to change during runtime so these are innocuous changes.

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
jilaada: Here is a thread which I had seen in the past go [over the distinction between node and clock](https://github.com/ros2/rclcpp/issues/595). So just to confirm I was under the impression that `rclcpp::Node::now()` will return the current ROS time rather than the current system time?

sensing/preprocessor/gnss/gnss_poser/CMakeLists.txt
```diff
@@ -1,66 +1,48 @@
...
+find_package(sensor_msgs)
+find_package(geometry_msgs)
+find_package(geo_pos_conv)
+find_package(ublox_msgs)
 
 find_package(PkgConfig)
-find_path(GeographicLib_INCLUDE_DIR GeographicLib/Config.h
-  PATH_SUFFIXES GeographicLib
-  )
+find_path(GeographicLib_INCLUDE_DIR GeographicLib/Config.h PATH_SUFFIXES GeographicLib)
```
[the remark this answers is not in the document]
jilaada: Thanks for the explanation, I think I should be able to get rid of these with new ament_cmake_auto calls!

sensing/preprocessor/gnss/gnss_poser/include/gnss_poser/convert.hpp
```diff
@@ -13,17 +13,20 @@
...
+#define GNSS_POSER_CONVERT_HPP_
 
 #include <GeographicLib/Geoid.hpp>
 #include <GeographicLib/MGRS.hpp>
 #include <GeographicLib/UTMUPS.hpp>
 
-#include "gnss_poser/gnss_stat.h"
-
 #include <geo_pos_conv/geo_pos_conv.hpp>
+#include <gnss_poser/gnss_stat.hpp>
```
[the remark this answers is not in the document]
jilaada: Ah yes, I meant to clean these up. Thanks for the explanation! I'll definitely make this fix, I actually ended up with an issue regarding header order when porting the multi_object_tracker package. Definitely lesson learnt!

sensing/preprocessor/gnss/gnss_poser/include/gnss_poser/convert.hpp
```diff
@@ -13,17 +13,20 @@
...
+
+#include <rclcpp/logging.hpp>
+#include <sensor_msgs/msg/nav_sat_fix.hpp>
 
-namespace GNSSPoser
+namespace
```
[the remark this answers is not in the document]
jilaada: Rename was needed because of an error with multiple definitions, I certainly agree its not good to have anonymous namespaces in a header. I wasn't sure how much refactoring was needed for this porting PR. But from my understanding this header file seems to provide wrapper methods for accessing and manipulating data related to the geographiclib library. Not sure you think it would be better to make these methods inline for this porting PR?

sensing/preprocessor/gnss/gnss_poser/include/gnss_poser/gnss_stat.hpp
```diff
@@ -13,10 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#pragma once
+#ifndef GNSS_POSER_GNSS_STAT_HPP_
+#define GNSS_POSER_GNSS_STAT_HPP_
 
-namespace GNSSPoser
```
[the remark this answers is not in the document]
jilaada: I'll add these back.

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
nnmm: That is correct, they return the same time. I'm fine with using it, it's just strange that you're calling it like a static function – how about `this->now()`?

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
nnmm: It is a lot nicer :) Since it's already here, no need to revert, but didn't we agree in the last standup to keep the component porting for later?

sensing/preprocessor/gnss/gnss_poser/package.xml
jilaada: We just need the geographiclib-tools package, as I had left the other exec depend in on accident while trying to get the code to compile and pull in the right resources. Thanks for catching that

jilaada:
@nnmm @fred-apex-ai Thanks for all the comments, I should have addressed them all in c539885, 8a81d18 and bd89bf8.

Once thing I am still not sure about is my change in bd89bf8 to use the node clock for time stamps, I followed [this](https://answers.ros.org/question/352364/ros2-throttle-logs/). Let me know if there are any conventions that might be out there already regarding throttle logs - I couldn't find an example in repo here.

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
```diff
@@ -13,76 +13,57 @@
...
+: rclcpp::Node("gnss_poser", node_options),
   tf2_listener_(tf2_buffer_),
-  coordinate_system_(CoordinateSystem::MGRS),
-  base_frame_("base_link"),
-  gnss_frame_("gnss"),
-  gnss_base_frame_("gnss_base_link"),
-  map_frame_("map"),
-  use_ublox_receiver_(false),
-  plane_zone_(9)
+  tf2_broadcaster_(this),
```
nnmm, line 27: Should this be `*this`?

sensing/preprocessor/gnss/gnss_poser/CMakeLists.txt
```diff
@@ -1,66 +1,48 @@
...
-  geometry_msgs
-  ublox_msgs
-  geo_pos_conv
-)
+### Compile options
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wno-unused-parameter -Wall -Wextra -Wpedantic -Werror)
```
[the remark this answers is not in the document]
fred-apex-ai: after today's standup, it has to be :-1: for `Werror` I am sorry :smile:

fred-apex-ai (approved):

fred-apex-ai:
@jilaada you probably just need to rebase to `ros2` because of the launch file renaming

sensing/preprocessor/gnss/gnss_poser/src/gnss_poser_core.cpp
jilaada: I couldn't find any references of people using `*this` in the ROS discourse though when doing some googling I did find that the geometry library tests were using `std::make_shared` and `shared_from_this()` so I changed the implementation to that:
* https://answers.ros.org/question/360516/ros2-tf2-broadcaster/ (implementation without `*this`)
* https://github.com/ros2/geometry2/blob/ros2/tf2_ros/test/test_transform_broadcaster.cpp#L46 (`shared_from_this()`)

jilaada:
@nnmm @fred-apex-ai I've rebased to `ros2`

mitsudome-r:
It looks like CI is failing when it tries to find ublox_msgs.
As mentioned in https://github.com/tier4/AutowareArchitectureProposal/pull/69, it is added to the build farm, but it would take a couple of weeks to become available.

@jilaada I think we can reopen the pull request , and build it from source until it becomes available.

jilaada:
@mitsudome-r I reopened the PR and made the necessary fixes for the CI to succeed.

jilaada:
@mitsudome-r @esteve It seems like the CI is still failing while trying to find the ublox_msgs dependencies. The PR that pulls the ublox source to AutowareArchitectureProposal has been merged but maybe that's not sufficient it seems - sorry, I haven't been following the CI discussion that has been ongoing, let me know if I need to make any changes anywhere.

mitsudome-r:
The ci should be solved after this PR is merged.
https://github.com/tier4/Pilot.Auto/pull/43

PR #20, opened 2020-10-12
Port ndt_omp to ROS2

nnmm:
* Removed ROS as a dependency, it was just used for its wall timer
* Replaced `pcl_ros` by `PCL`
* See comments for CMake questions
* There is a large number of warnings from using deprecated `pcl` functions. Maybe this already has been fixed in the newest version of https://github.com/koide3/ndt_omp?

### DIFF
--- a/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
+++ b/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
@@ -27,6 +27,7 @@ endif()
 
 include_directories(
   include
+  ${PCL_COMMON_INCLUDE_DIRS}
 )
 
 add_library(ndt_omp
@@ -52,13 +53,13 @@ add_executable(align
   apps/align.cpp
 )
 
-target_link_libraries(align
-  rclcpp
+# get_target_property(FRED_PCL_LINK_LIBRARIES PCL::pcl LINK_INTERFACE_LIBRARIES)
+
+# ament_target_dependencies can't depend on targets, i.e. ndt_omp
+target_link_libraries(align PUBLIC
   ndt_omp
-  PCL
+  ${PCL_LIBRARIES}
 )
-
-
 
 install(TARGETS ndt_omp
   DESTINATION lib/${PROJECT_NAME}
@@ -66,7 +67,7 @@ install(TARGETS ndt_omp
 
 ## Install project namespaced headers
 install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION share/${PROJECT_NAME}
+  DESTINATION install/${PROJECT_NAME}
 )
 
 ament_package()
--- a/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
+++ b/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
@@ -1,16 +1,14 @@
 cmake_minimum_required(VERSION 3.5)
 project(ndt_omp)
 
-# -mavx causes a lot of errors!!
-add_definitions(-std=c++14 -msse -msse2 -msse3 -msse4 -msse4.1 -msse4.2)
-set(CMAKE_CXX_FLAGS "-std=c++14 -msse -msse2 -msse3 -msse4 -msse4.1 -msse4.2")
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
 
 # warn performance issue of ndt_omp which is NOT built in release mode
-# string(TOUPPER ${CMAKE_BUILD_TYPE} uppercase_CMAKE_BUILD_TYPE)
-# if (NOT (${uppercase_CMAKE_BUILD_TYPE} STREQUAL "RELEASE"))
-#   message(WARNING
-#     "CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} may cause a serious performance degradation. Please configure with -DCMAKE_BUILD_TYPE=Release.")
-# endif()
+if (NOT ("${CMAKE_BUILD_TYPE}" STREQUAL "Release"))
+  message(WARNING "CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} may cause a serious performance degradation. Please configure with -DCMAKE_BUILD_TYPE=Release.")
+endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_auto REQUIRED)
@@ -38,8 +36,12 @@ target_include_directories(ndt_omp
 )
 
 # pcl 1.7 causes a segfault without -O2 (or -O3) flag
-target_compile_options(ndt_omp PUBLIC
-  $<$<CONFIG:Debug>:-O2 -g>
+target_compile_options(ndt_omp
+  PUBLIC
+    $<$<CONFIG:Debug>:-O2 -g>
+  PRIVATE
+    # -mavx causes a lot of errors!!
+    -msse -msse2 -msse3 -msse4 -msse4.1 -msse4.2
 )
 
 set(NDT_OMP_DEPENDENCIES
--- a/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
+++ b/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
@@ -54,6 +54,6 @@ add_executable(align
   apps/align.cpp
 )
 
-ament_target_dependencies(ndt_omp ${NDT_OMP_DEPENDENCIES}) 
+target_link_libraries(align ndt_omp ${PCL_LIBRARIES})
 
 ament_auto_package()
--- a/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
+++ b/localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(ndt_omp)
 
 # -mavx causes a lot of errors!!
@@ -6,28 +6,20 @@ add_definitions(-std=c++14 -msse -msse2 -msse3 -msse4 -msse4.1 -msse4.2)
 set(CMAKE_CXX_FLAGS "-std=c++14 -msse -msse2 -msse3 -msse4 -msse4.1 -msse4.2")
 
 # warn performance issue of ndt_omp which is NOT built in release mode
-string(TOUPPER ${CMAKE_BUILD_TYPE} uppercase_CMAKE_BUILD_TYPE)
-if (NOT (${uppercase_CMAKE_BUILD_TYPE} STREQUAL "RELEASE"))
-  message(WARNING
-    "CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} may cause a serious performance degradation. Please configure with -DCMAKE_BUILD_TYPE=Release.")
-endif()
+# string(TOUPPER ${CMAKE_BUILD_TYPE} uppercase_CMAKE_BUILD_TYPE)
+# if (NOT (${uppercase_CMAKE_BUILD_TYPE} STREQUAL "RELEASE"))
+#   message(WARNING
+#     "CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} may cause a serious performance degradation. Please configure with -DCMAKE_BUILD_TYPE=Release.")
+# endif()
 
-find_package(catkin REQUIRED COMPONENTS
-  roscpp
-  pcl_ros
-)
-
+find_package(ament_cmake REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(PCL REQUIRED)
 find_package(OpenMP)
 if (OPENMP_FOUND)
   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
 endif()
-
-catkin_package(
-  INCLUDE_DIRS include
-  LIBRARIES ndt_omp
-  CATKIN_DEPENDS
-)
 
 ###########
 ## Build ##
@@ -35,7 +27,6 @@ catkin_package(
 
 include_directories(
   include
-  ${catkin_INCLUDE_DIRS}
 )
 
 add_library(ndt_omp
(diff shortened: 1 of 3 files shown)


localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,61 @@
...
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
 
 # warn performance issue of ndt_omp which is NOT built in release mode
-string(TOUPPER ${CMAKE_BUILD_TYPE} uppercase_CMAKE_BUILD_TYPE)
-if (NOT (${uppercase_CMAKE_BUILD_TYPE} STREQUAL "RELEASE"))
-  message(WARNING
-    "CMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} may cause a serious performance degradation. Please configure with -DCMAKE_BUILD_TYPE=Release.")
+if (NOT ("${CMAKE_BUILD_TYPE}" STREQUAL "Release"))
```
nnmm, line 9: This will always emit a warning since I think we rarely compile in Release mode. Is this the desired behavior?
How about just always enabling optimization with an unconditional `-O2` flag or similar?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,61 @@
...
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
+ament_target_dependencies(ndt_omp ${NDT_OMP_DEPENDENCIES}) 
+
+
+add_executable(align
+  apps/align.cpp
 )
+
+target_link_libraries(align ndt_omp ${PCL_LIBRARIES})
```
nnmm, line 59: I have to write `${PCL_LIBRARIES}` here, `ament_target_dependencies(align ${NDT_OMP_DEPENDENCIES})` doesn't work even though it appears to work in https://github.com/tier4/Pilot.Auto/pull/21/files. I would prefer to use `ament_target_dependencies` since it also takes `PCL_LIBRARY_DIRS`, `PCL_INCLUDE_DIRS` etc. into account and is shorter.

@fred-apex-ai Any idea why?

nnmm:
@simon-t4 Sorry for updating after I assigned you, but now I think it is ready for review.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
fred-apex-ai: @nnmm I guess you could set the release type yourself. But this check is an anti-pattern anyways as it wouldn't work the multi-config builds such as with clion; see https://stackoverflow.com/a/24470998/14409577

But yes, this highlights an important point that I raised here https://github.com/colcon/colcon-cmake/issues/96. For the next standup, we should bring up the default build config

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
fred-apex-ai: @nnmm did our discussion solve this issue?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,87 @@
...
-)
-add_dependencies(align
-  ndt_omp
+target_include_directories(ndt_omp
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include> 
 )
+
 # pcl 1.7 causes a segfault without -O2 (or -O3) flag
```
simon-t4, line 37: Just a comment - the default installation of pcl for Ubuntu 20.04 is 1.10. Are the compile options still necessary? I guess you do more here with the msse options though so easier to leave as is?

Also for me installing pcl required the package libpcl-dev, but I see below you install libpcl-all-dev, I think I remember reading about this issue, but if you need libpcl-all-dev and had to do something to find it, can you please add in these comments for a record.

nnmm: That's a good question that's difficult to answer. The comment doesn't say which part segfaults, or whether PCL segfaults immediately, so it's hard to reproduce, but I'd expect such a serious bug to be fixed by now. There seems to be no open issue about this either on the PCL project's GitHub. I'm fine with removing it and if PCL segfaults, we can check back.

nnmm: I think `libpcl-all-dev` is the "ROS name" (not sure what the name for that abstraction layer is) for the `libpcl-dev` package on Ubuntu: https://github.com/ros/rosdistro/blob/f8335c214e20d445756ba6479e69a050bbeb3094/rosdep/base.yaml#L3429

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
nnmm: I just set the build type to Release, that's what the upstream packages does too.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
nnmm: I tried out the `align` binary and it doesn't segfault even when the build type is not set to Release, so it should be fine.

nnmm:
@simon-t4 could you take another look?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,81 @@
...
-)
+set(CMAKE_BUILD_TYPE "Release")
 
+find_package(ament_cmake REQUIRED)
+find_package(PCL REQUIRED COMPONENTS common filters kdtree registration io)
 find_package(OpenMP)
+
 if (OPENMP_FOUND)
   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
   set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
```
fred-apex-ai, line 16: I'm surprised that it works w/o updating linker flags as in here. Note that I wrote this 3 years ago

https://github.com/mtazzari/galario/blob/master/src/CMakeLists.txt#L63

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,81 @@
...
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include> 
 )
-add_dependencies(align
-  ndt_omp
+
+target_compile_options(ndt_omp
+  PRIVATE
+    # -mavx causes a lot of errors!!
+    -msse -msse2 -msse3 -msse4 -msse4.1 -msse4.2
```
fred-apex-ai, line 37: addings these flags is not portable. Should wrap in 
```
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang") 
```

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,81 @@
...
+add_executable(align
+  apps/align.cpp
 )
 
-install(TARGETS ndt_omp
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+# Strangely, for an executable, ament_target_dependencies(align ${NDT_OMP_DEPENDENCIES})
+# gives a linker error from PCL
```
fred-apex-ai, line 53: I don't understand this comment, or what it's consequences are. Please clarify

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,81 @@
...
+ament_target_dependencies(align PUBLIC OpenMP)
+# So we include PCL manually
+target_include_directories(align PUBLIC ${PCL_INCLUDE_DIRS})
+target_link_libraries(align PUBLIC ${PCL_LIBRARIES})
```
fred-apex-ai, line 57: [nit] I'd regroup to have the compile and link commands together
```
target_compile_definitions
target_include_directories
target_link_libraries
...
```
``

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
```diff
@@ -1,70 +1,81 @@
...
+install(
+  TARGETS ndt_omp
+  EXPORT export_ndt_omp
+  LIBRARY DESTINATION lib
+  ARCHIVE DESTINATION lib
+  RUNTIME DESTINATION bin
+  INCLUDES DESTINATION include
 )
+
+ament_package()
```
fred-apex-ai, line 80: can't you do `ament_package_auto()` instead of the long `install(...)` commands?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/apps/align.cpp
```diff
@@ -45,17 +44,16 @@ pcl::PointCloud<pcl::PointXYZ>::Ptr align(
...
+  auto t2 = std::chrono::steady_clock::now();
+  std::cout << "single : " << std::chrono::duration_cast<std::chrono::milliseconds>(t2 - t1).count() << "[msec]" << std::endl;
 
   for (int i = 0; i < 10; i++) {
     registration->align(*aligned);
   }
-  auto t3 = ros::WallTime::now();
-  std::cout << "10times: " << (t3 - t2).toSec() * 1000 << "[msec]" << std::endl;
+  auto t3 = std::chrono::steady_clock::now();
+  std::cout << "10times: " << std::chrono::duration_cast<std::chrono::milliseconds>(t3 - t2).count() << "[msec]" << std::endl;
```
fred-apex-ai, line 56: We should use the logging facilities instead of `cout`

localization/pose_estimator/ndt_scan_matcher/ndt_omp/package.xml
```diff
@@ -1,5 +1,6 @@
 <?xml version="1.0"?>
-<package>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
```
fred-apex-ai, line 2: I have seen and preserved 
```

<?xml version="1.0"?>
<package format="2">
```
elsewhere. Are you following the recommendations and I don't? What are the implications?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/apps/align.cpp
```diff
@@ -45,17 +44,16 @@ pcl::PointCloud<pcl::PointXYZ>::Ptr align(
...
+  auto t1 = std::chrono::steady_clock::now();
   registration->align(*aligned);
-  auto t2 = ros::WallTime::now();
-  std::cout << "single : " << (t2 - t1).toSec() * 1000 << "[msec]" << std::endl;
+  auto t2 = std::chrono::steady_clock::now();
+  std::cout << "single : " << std::chrono::duration_cast<std::chrono::milliseconds>(t2 - t1).count() << "[msec]" << std::endl;
 
   for (int i = 0; i < 10; i++) {
     registration->align(*aligned);
   }
```
simon-t4, line 54: Maybe outside the scope of porting - but this is strange. Call align once, then call ten more times? Is this setting up the parallelism by calling 10 times, or is it simple meant to iteratively provide a better registration?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/package.xml
simon-t4: I have seen both package format 2 & 3 as well. 
From here 
https://www.ros.org/reps/rep-0149.html
format 3 seems to enable mixing ros1 ad ros2 builds.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/apps/align.cpp
simon-t4: @nnmm Also if output is necessary perhaps should combine output with previous cout (single and 10x's time in one output/log)?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
nnmm: I'm wary of mixing the two. I've tried it before and ran into problems.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/apps/align.cpp
nnmm: I think it might be some sort of benchmarking.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/apps/align.cpp
nnmm: @fred-apex-ai This code is independent from rclcpp, why would `std::cout` not be suitable here?

localization/pose_estimator/ndt_scan_matcher/ndt_omp/package.xml
nnmm: The migration guide says "ROS 2 doesn’t support format 1 of the package specification but only newer format versions (2 and higher)." My understanding is that if you have a package that's already at version 2 it's fine to leave it, but this one was at version 1, and going straight to 3 doesn't hurt.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
nnmm: Done.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
nnmm: Changed to a more modern style of using OpenMP.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/CMakeLists.txt
nnmm: Changed to a more modern style of using OpenMP.

nnmm:
@simon-t4 Apologies for yet another iteration on CMakeLists.txt – I discovered that downstream packages would have problems. Now it all works and I documented why it is the way it is.

Could you review & merge (if you find no issues in your review) this soon since it's blocking two dependent packages? I'd prefer to steer away from most functional changes in this PR, especially align.cpp which might not even be used anywhere.

localization/pose_estimator/ndt_scan_matcher/ndt_omp/apps/align.cpp
simon-t4: @nnmm I agree - and don't think this is used very much. As you say below outside the scope of this porting.

simon-t4 (approved):

nnmm:
Awesome, thanks @simon-t4!

PR #21, opened 2020-10-12
Port ndt_pcl_modified to ROS2

nnmm:
* I tried with `ament_auto` but it didn't find PCL.


nnmm:
Please review, @jilaada!

localization/pose_estimator/ndt_scan_matcher/ndt_pcl_modified/CMakeLists.txt
```diff
@@ -1,15 +1,17 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(ndt_pcl_modified)
 
-add_compile_options(-std=c++14)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
```
jilaada, line 8: Minor: I think there was some talk about adding `-Werror` as well: https://github.com/tier4/Pilot.Auto/pull/6#discussion_r503324085

jilaada (commented):
I'm not sure how to test this since this a library, though I assume you might have already needed this package as a dependency for another package you're working on. I do have some concern though regarding the Cmakelist, since this is a library I think in Foxy there is some change in the way include directories are exported.

>  For library targets to be exportable like this they must not rely on classic functions affecting global state like include_directories() but set the include directories on the target itself - for the build as well as install environment - using generator expressions, e.g. target_include_directories(<target> PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>" "$<INSTALL_INTERFACE:include>")

This was taken from https://index.ros.org/doc/ros2/Releases/Release-Foxy-Fitzroy/#id14. 

I had this issue when porting the `geo_pos_conv` package as its header could not be found by the `gnss_poser` package. Once again not sure if its applied here.

localization/pose_estimator/ndt_scan_matcher/ndt_pcl_modified/CMakeLists.txt
nnmm: Done

nnmm:
@jilaada That's a good point, I wasn't quite happy with the CMakeLists.txt format as well. I now used `ament_auto`'s `ament_auto_package()` which should do the job of `ament_export_include_directories()` mentioned in the linked article, and used generator expressions for `target_include_directories()`. That might still not be the cleanest way, but it's the best way I could find to handle PCL for now.

It is "tested" by being included in https://github.com/tier4/Pilot.Auto/pull/29, which compiles successfully for me.

nnmm:
@jilaada Updated the CMakeLists.txt yet again, sorry – I think now I'm happy with it though.

jilaada:
@nnmm Out of curiosity, what was the reason for reverting `ament_cmake_auto`?

nnmm:
I'm on mobile and can't link to it, but see PR 29

jilaada:
> I'm on mobile and can't link to it, but see PR 29

For some reason I didn't get a notification for this! Else I would have approved earlier. I just checked, and re-tested the package. LGTM :+1:

jilaada (approved):

nnmm:
Don't worry about not approving immediately, it must have been the middle of the night for you! I certainly don't want to disturb you guys outside working hours.

nnmm:
However, could you please merge the PR? I don't think I have permissions for that.

jilaada:
No worries! I've merged the PR, thanks!

PR #24, opened 2020-10-13
ROS2 Porting: multi_object_tracker

jilaada:
# multi_object_tracker

## Summary

Convert the multi_object_tracker package to ROS2. A large but straightforward change with a few caveats which is why this PR is currently a draft.

Important (API) Changes:
* Convert `DataAssociation::assign` method return type to void due to `-Wreturn-type` error
* Pass `rclcpp::Time` by reference to `TrackerBase::getElapsedTimeFromLastUpdate` to perform duration calculation

Cosmetic Changes:
* Adding `#ifndef` guards on all header files and removing `#pragma once`
* Standardise use of `<>` vs `""` in `#include`
* Convert `.h` extension to `.hpp`
* Typos

## Testing

1. Checkout this branch and enter docker/ade environment

2. Build and source the repository
```
colcon build
. install/setup.bash
```
3. Launch file should launch without error
```
ros2 launch multi_object_tracker multi_object_tracker.launch.xml
```

### `lookupTransform`
The method `lookupTransform` in `tf2::Buffercore` no longer allows specification of lookup duration. At the moment I have left this implement to perform the lookup transform without a duration specification, see [issue here for further discussion](https://github.com/tier4/Pilot.Auto/issues/23).

**Update:** This has been implemented using the suggestion provided in the issue and taking inspiration from https://github.com/ros2/geometry2/blob/foxy/tf2_ros/test/test_buffer.cpp#L140-L158.

### `uuid_msgs` and `unique_id` Packages
This packages contains these dependencies however they have been ported (though the state of port is unclear) but not released on the ROS2 package index. I have commented on an issue tracking its progress regarding any further porting efforts.

**Update:** This has been implemented using the `unique_identifier_msgs` package which has a release available on foxy. I removed any references to Boost, using https://github.com/ros2/unique_identifier_msgs/blob/foxy/msg/UUID.msg instead. A different method of generate UUIDs using standard libraries was implemented instead of using the generate method in the `unique_id` package. Some references:
* https://github.com/ros2/rclcpp/blob/31c202e325476f6e912d6d23c17d13393156dbf4/rclcpp_action/src/client.cpp#L377-L379
* https://lowrey.me/guid-generation-in-c-11/


perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/vehicle_tracker.cpp
```diff
@@ -205,13 +207,14 @@ bool VehicleTracker::measure(
 }
 
 bool VehicleTracker::getEstimatedDynamicObject(
-  const ros::Time & time, autoware_perception_msgs::DynamicObject & object)
+  const rclcpp::Time & time, autoware_perception_msgs::msg::DynamicObject & object)
 {
   object = object_;
-  object.id = unique_id::toMsg(getUUID());
+  // TEMP to replace with the uuid generation
+  // object.id = unique_id::toMsg(getUUID());
```
mitsudome-r, line 214: Since we solved the generation of uuid, we can now uncomment this correct?
```suggestion
  object.id = getUUID();
```

(and the same for all other places that use getUUID(). )

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
```diff
@@ -38,20 +33,28 @@ add_executable(multi_object_tracker_node
...
+  src/data_association/successive_shortest_path/successive_shortest_path.hpp
+  include/multi_object_tracker/data_association/data_association.hpp
+  include/multi_object_tracker/tracker/tracker.hpp
+  include/multi_object_tracker/tracker/model/tracker_base.hpp
+  include/multi_object_tracker/tracker/model/bicycle_tracker.hpp
+  include/multi_object_tracker/tracker/model/pedestrian_tracker.hpp
+  include/multi_object_tracker/tracker/model/vehicle_tracker.hpp
+  include/multi_object_tracker/utils/utils.hpp
+  include/multi_object_tracker/multi_object_tracker_core.hpp
 )
```
mitsudome-r, line 46: ament_cmake_auto should add include file automatically, so you shouldn't have to set this.

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
```diff
@@ -38,20 +33,28 @@ add_executable(multi_object_tracker_node
...
+  ${MULTI_OBJECT_TRACKER_HEADERS}
+)
 
-install(
-  DIRECTORY
-    launch
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
+rclcpp_components_register_node(multi_object_tracker_node
+  PLUGIN "MultiObjectTracker"
+  EXECUTABLE multi_object_tracker
```
mitsudome-r, line 55: We decided not to convert to rclcpp_components at initial porting.
We don't have to revert this PR, but when you port other packages, you don't have to make them into component nodes.

perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/vehicle_tracker.cpp
jilaada: Ah, good catch 9d3c8a1

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
jilaada: I was told in a different PR that this is better practice: https://github.com/tier4/Pilot.Auto/pull/18#discussion_r504575682. I figured it applied here too, maybe I'm wrong.

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
jilaada: Yeah, sorry this was the last PR with `rclcpp_components`. This package was ported a while ago when we were in limbo about the usage of components. Subsequent packages use the hand-written main file.

nnmm:
Apologies @jilaada, I missed when this PR switched from draft to ready for review. I will review it today.

jilaada:
> Apologies @jilaada, I missed when this PR switched from draft to ready for review. I will review it today.

No problem, I'm still looking into the transform issue anyway.

fred-apex-ai:
> > Apologies @jilaada, I missed when this PR switched from draft to ready for review. I will review it today.
> 
> No problem, I'm still looking into the transform issue anyway.

@jilaada Are you looking for something like this https://github.com/fred-apex-ai/autoware.iv.universe/commit/32f75580b6c6c445d2424261a7fc7726f3db3e75#diff-30dcb55e06b0d11249fb5d74414c4c96ceb4b891a18142afed00452ed9ef2ec3R145?

jilaada:
> > No problem, I'm still looking into the transform issue anyway.
> 
> @jilaada Are you looking for something like this [fred-apex-ai@32f7558#diff-30dcb55e06b0d11249fb5d74414c4c96ceb4b891a18142afed00452ed9ef2ec3R145](https://github.com/fred-apex-ai/autoware.iv.universe/commit/32f75580b6c6c445d2424261a7fc7726f3db3e75#diff-30dcb55e06b0d11249fb5d74414c4c96ceb4b891a18142afed00452ed9ef2ec3R145)?

Actually quite similar. In your case you don't need to propagate the transform and use it, you're just waiting to receive the transform to continue the execution process. In my case, I am trying to find an equivalent to a lookupTransform where the duration is set and the execution cycle is blocked until the transform is received. Here is what I have currently: https://github.com/tier4/Pilot.Auto/pull/24/files#diff-1769717c3f36f32c9c336458b8968ddb2285a9a7794ce8004f308a0892783633R77-R88

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+#include <string>
+
+MultiObjectTracker::MultiObjectTracker(const rclcpp::NodeOptions & node_options)
+: rclcpp::Node("multi_object_tracker", node_options)
+{
+  // Create publishers and subscribers
+  dynamic_object_sub_ =
+    create_subscription<autoware_perception_msgs::msg::DynamicObjectWithFeatureArray>(
+      "input", rclcpp::QoS{1},
```
nnmm, line 37: This had `latch = true` before, if this behavior should be kept, this would need the `transient_local` durability QoS.

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -0,0 +1,200 @@
...
+      std::bind(&MultiObjectTracker::measurementCallback, this, std::placeholders::_1));
+  dynamic_object_pub_ =
+    create_publisher<autoware_perception_msgs::msg::DynamicObjectArray>("output", rclcpp::QoS{1});
+
+  // Parameters
+  double publish_rate = declare_parameter<double>("publish_rate", 30.0);
+  world_frame_id_ = declare_parameter<std::string>("world_frame_id", std::string("world"));
```
nnmm, line 44: I think the two explicit `std::string` are not needed (but it's fine to leave it in if you prefer).

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+  rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
+  tf_buffer_ = std::make_shared<tf2_ros::Buffer>(clock);
```
nnmm, line 47: I would initialize this and `tf_listener_` in the initializer list, with `this->get_clock()`.

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -0,0 +1,200 @@
...
+      tf_world2objects = tf_world2objects_world * tf_objects_world2objects;
+      tf2::toMsg(
+        tf_world2objects,
+        input_transformed_objects.feature_objects.at(i).object.state.pose_covariance.pose);
+    }
+  }
+
+  /* tracker prediction */
+  rclcpp::Time measurement_time = input_objects_msg->header.stamp;
+  for (auto itr = list_tracker_.begin(); itr != list_tracker_.end(); ++itr) {
```
nnmm, line 106: Really surprised that this isn't a range-based for. I find range-based for much easier to read. I know it's a style issue though, so feel free to leave it like it is.

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -0,0 +1,200 @@
...
+  /* global nearest neighboor */
+  std::unordered_map<int, int> direct_assignment;
+  std::unordered_map<int, int> reverse_assignment;
+  Eigen::MatrixXd score_matrix = data_association_.calcScoreMatrix(
+    input_transformed_objects, list_tracker_);  // row : tracker, col : measurement
+  data_association_.assign(score_matrix, direct_assignment, reverse_assignment);
+
+  /* tracker measurement update */
+  int tracker_idx = 0;
+  for (auto tracker_itr = list_tracker_.begin(); tracker_itr != list_tracker_.end();
```
nnmm, line 122: If you choose to fix for loops, here is another one

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
```diff
@@ -0,0 +1,200 @@
...
+      list_tracker_.erase(erase_itr);
+    }
+  }
+
+  // Create output msg
+  rclcpp::Time current_time = rclcpp::Node::now();
+  autoware_perception_msgs::msg::DynamicObjectArray output_msg;
+  output_msg.header.frame_id = world_frame_id_;
+  output_msg.header.stamp = current_time;
+  for (auto itr = list_tracker_.begin(); itr != list_tracker_.end(); ++itr) {
```
nnmm, line 188: If you choose to fix for loops, here is another one

perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/tracker_base.cpp
```diff
@@ -20,17 +20,27 @@
...
+
+Tracker::Tracker(const rclcpp::Time & time, const int type)
+: type_(type),
   no_measurement_count_(0),
   total_measurement_count_(1),
   last_update_with_measurement_time_(time)
 {
+  // Generate random number
+  std::generate(uuid_.uuid.begin(), uuid_.uuid.end(), [] {
+    std::random_device rd;
```
nnmm, line 34: I think it is more idiomatic to initialize these outside `std::generate`, and a uniform distribution is fine too but I find `std::independent_bits_engine` a bit more concise:
```
std::mt19937 rng(std::random_device{}());
std::independent_bits_engine<std::mt19937, 8, uint8_t> byte_rng(rng);
std::generate(uuid_.uuid.begin(), uuid_.uuid.end(), byte_rng);
```

nnmm (approved):
Looks great! For the waitForTransform, see the discussion in the issue.

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
```diff
@@ -38,20 +33,28 @@ add_executable(multi_object_tracker_node
...
+  include/multi_object_tracker/multi_object_tracker_core.hpp
 )
 
-target_link_libraries(multi_object_tracker_node ${catkin_LIBRARIES})
-
-install(TARGETS multi_object_tracker_node
-        ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-        RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION})
+ament_auto_add_library(multi_object_tracker_node SHARED
```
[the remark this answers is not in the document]
jilaada: This shared was required for the node to be able to be launch as a component in the launch file. See https://github.com/tier4/Pilot.Auto/pull/6#issuecomment-708254090

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
```diff
@@ -1,34 +1,29 @@
...
-    tf2
-    tf2_ros
-    unique_id
-    uuid_msgs
-)
+ament_auto_find_build_dependencies()
 
-include_directories(include ${catkin_INCLUDE_DIRS} ${EIGEN3_INCLUDE_DIR})
+# includes
+include_directories(SYSTEM ${EIGEN3_INCLUDE_DIR})
```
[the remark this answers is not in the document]
jilaada: Without this line I was not able to find the eigen headers, the error output below is returned when that line is removed:

```
jilada@ade:~/AutowareArchitectureProposal (ros2-porting/add-ublox-deps $% u=)$ colcon build --packages-up-to multi_object_tracker
Starting >>> autoware_perception_msgs
Finished <<< autoware_perception_msgs [0.58s]
Starting >>> multi_object_tracker
--- stderr: multi_object_tracker
In file included from /home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/include/multi_object_tracker/multi_object_tracker_core.hpp:22,
                 from /home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp:20:
/home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/include/multi_object_tracker/data_association/data_association.hpp:26:10: fatal error: Eigen/Core: No such file or directory
   26 | #include <Eigen/Core>
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/multi_object_tracker_node.dir/build.make:63: CMakeFiles/multi_object_tracker_node.dir/src/multi_object_tracker_core.cpp.o] Error 1
make[2]: *** Waiting for unfinished jobs....
/home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/bicycle_tracker.cpp:28:10: fatal error: Eigen/Core: No such file or directory
   28 | #include <Eigen/Core>
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/multi_object_tracker_node.dir/build.make:115: CMakeFiles/multi_object_tracker_node.dir/src/tracker/model/bicycle_tracker.cpp.o] Error 1
In file included from /home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/src/data_association/data_association.cpp:20:
/home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/include/multi_object_tracker/data_association/data_association.hpp:26:10: fatal error: Eigen/Core: No such file or directory
   26 | #include <Eigen/Core>
      |          ^~~~~~~~~~~~
compilation terminated.
/home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/pedestrian_tracker.cpp:28:10: fatal error: Eigen/Core: No such file or directory
   28 | #include <Eigen/Core>
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/multi_object_tracker_node.dir/build.make:141: CMakeFiles/multi_object_tracker_node.dir/src/data_association/data_association.cpp.o] Error 1
make[2]: *** [CMakeFiles/multi_object_tracker_node.dir/build.make:128: CMakeFiles/multi_object_tracker_node.dir/src/tracker/model/pedestrian_tracker.cpp.o] Error 1
/home/jilada/AutowareArchitectureProposal/src/autoware/autoware.iv/perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/vehicle_tracker.cpp:28:10: fatal error: Eigen/Core: No such file or directory
   28 | #include <Eigen/Core>
      |          ^~~~~~~~~~~~
compilation terminated.
make[2]: *** [CMakeFiles/multi_object_tracker_node.dir/build.make:102: CMakeFiles/multi_object_tracker_node.dir/src/tracker/model/vehicle_tracker.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:107: CMakeFiles/multi_object_tracker_node.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make: *** [Makefile:141: all] Error 2
---
Failed   <<< multi_object_tracker [3.59s, exited with code 2]

```

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
```diff
@@ -1,34 +1,29 @@
...
-  tf2_ros
-  unique_id
-  uuid_msgs
-)
+### Find Packages
+find_package(ament_cmake_auto REQUIRED)
 
-find_package(Eigen3 REQUIRED)
+## Find Eigen Dependencies
+find_package(eigen3_cmake_module REQUIRED)
```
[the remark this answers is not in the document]
jilaada: When I remove the `find_package(eigen_cmake_module)` and the `find_package(Eigen3)` calls, I cannot find the Eigen headers. I call the `ament_auto_find_build_dependencies()` call after these because I thought it grouped the calls a little better, though I see how it looks quite confusing I'll change this.

Are there other packages which have used the Eigen libraries and could just depend on the `ament_auto_find_build_dependencies()` call to find the package?

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
jilaada: Ah good catch, I missed this. Changed in 0febc64

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
jilaada: Good point, changed in 0febc64

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
jilaada: Ah yes, I have also made this plain objects in addition to moving it in the intializer list here: 0febc64

perception/object_recognition/tracking/multi_object_tracker/src/tracker/model/tracker_base.cpp
jilaada: Thanks for the suggestion, I don't know too much about this and was mostly following the way it was computed in the `unique_id` package though I guess this might/would change when that package is ported. I've taken your suggestion into account here: 0febc64

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
jilaada: I totally agree with you, I didn't find these all that easy to read either. As far as I'm concerned there are plenty of other places and packages which could benefit from these stylistic changes. Maybe we could create an issue for this and append any other stylistic changes to work on there? If you're ok with this I can create the issue

jilaada:
@nnmm I have addressed your PR comments. Please take a look at my changes again when you get a chance. Regarding the Eigen issue; I have the minimum set of CMake commands to find the Eigen library. Removing any of these will cause the error output I commented earlier. Regarding the for loops, see my comment here: https://github.com/tier4/Pilot.Auto/pull/24#discussion_r512392210

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
nnmm: Sure, I'm ok with an issue

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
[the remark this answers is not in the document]
nnmm: I see, thanks!

perception/object_recognition/tracking/multi_object_tracker/src/multi_object_tracker_core.cpp
jilaada: Here is the issue: https://github.com/tier4/Pilot.Auto/issues/61

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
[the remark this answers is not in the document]
nnmm: Does `target_link_libraries(my_target Eigen3::Eigen)` like described [here](https://index.ros.org/doc/ros2/Tutorials/Ament-CMake-Documentation/#adding-dependencies) instead of `
include_directories(SYSTEM ${EIGEN3_INCLUDE_DIR})` work?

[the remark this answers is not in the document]
jilaada: That works excellently, I just push my changes: 0b8b53d

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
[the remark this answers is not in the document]
fred-apex-ai: @jilaada I used it [here](https://github.com/tier4/Pilot.Auto/blob/ros2/common/math/osqp_interface/CMakeLists.txt) w/o issues. Could update that to use `Eigen3::Eigen` as @nnmm recommended but otherwise it's equivalent to your context here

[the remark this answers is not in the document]
jilaada: Thanks for the information. I've changed it according to @nnmm suggestion in this comment: https://github.com/tier4/Pilot.Auto/pull/24#discussion_r512511753. Works excellently!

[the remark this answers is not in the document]
nnmm: You're probably aware of this yourself, but just a side note: I wouldn't go too far with "if it works and pulls in the least number of dependencies, it's correct" – I often noticed that colcon does not complain if you leave out theoretically-necessary dependencies in package.xml and CMakeLists.txt. Those packages are then subtly broken and only work by accident, or will reveal problems in packages that depend on them.

So even if your package would work only with `ament_auto_find_build_dependencies()`, I'd think the way you did it here is correct since [I wouldn't assume `ament_auto_find_build_dependencies()` will `find_package()` in any particular order](https://github.com/ament/ament_cmake/blob/f123edf40390be5bcfa015105c76ccd763acac3d/ament_cmake_auto/cmake/ament_auto_find_build_dependencies.cmake#L62) and [the `eigen3_cmake_module` readme says to find `Eigen3` after `eigen3_cmake_module`](https://github.com/ros2/eigen3_cmake_module#edit-your-cmakeliststxt).

nnmm (approved):
Great, ready to merge from my side!

perception/object_recognition/tracking/multi_object_tracker/CMakeLists.txt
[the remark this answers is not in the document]
jilaada: The second doc you linked was what I used as reference to add the eigen dependencies. I thought the order in which packages are added doesn't matter - but now that I know I'll keep that in mind for the future. I'll add this find_package call back. Added back here: 160b275

PR #26, opened 2020-10-13
Fix simple planning simulator

TakaHoribe:
- Fix launch
 - Fix timer callback
 - Add default param

How to test

Launch simulator, 

```
ros2 launch simple_planning_simulator simple_planning_simulator.launch.xml
```

then, send `/initialpose` from Rviz. Check if the `tf` is published correctly.


mitsudome-r (approved):
LGTM
I tested launch file and confirmed that the node publishes tf after setting initial pose.

PR #27, opened 2020-10-13
port velocity controller

TakaHoribe:
TODO
 - [x] port main functions
 - [x] port dynamic_reconfigure-functions

The `dynamic reconfigure` function is ported as the service callback. 

The callback is set like this.
```
set_param_res_ =
  this->add_on_set_parameters_callback(std::bind(&VelocityController::paramCallback, this, _1));
```

The parameters are set using the `set_double_param_if_changed()` function defined locally in the `paramCallback()`.


TakaHoribe:
Still need to wait to merge for the timeout issue of tf_listener

TakaHoribe:
@fred-apex-ai Could you review this?

control/velocity_controller/CMakeLists.txt
```diff
@@ -1,77 +1,23 @@
...
-include_directories(
-  include
-  ${catkin_INCLUDE_DIRS}
-)
 
-add_executable(
-  velocity_controller
+ament_auto_add_executable(velocity_controller
   src/velocity_controller_core.cpp
   src/velocity_controller_node.cpp
```
fred-apex-ai, line 14: please also add the headers. This will help IDEs to discover which source file are part of the project. Won't change the executable that's produced

control/velocity_controller/include/velocity_controller/delay_compensation.h
```diff
@@ -17,7 +17,7 @@
 #ifndef VELOCITY_CONTROLLER_DELAY_COMPENSATION
 #define VELOCITY_CONTROLLER_DELAY_COMPENSATION
 
-#include <autoware_planning_msgs/Trajectory.h>
+#include <autoware_planning_msgs/msg/trajectory.hpp>
```
fred-apex-ai, line 20: please reverse header order so it's from local to other pkgs to std includes

control/velocity_controller/include/velocity_controller/delay_compensation.h
```diff
@@ -39,12 +39,13 @@ class DelayCompensator
     auto pred_pose = current_pose;
     pred_pose.pose.position.x += dx;
     pred_pose.pose.position.y += dy;
-    pred_pose.header.stamp += ros::Duration(delay_time);
+    pred_pose.header.stamp =
+      rclcpp::Time(pred_pose.header.stamp) + rclcpp::Duration::from_seconds(delay_time);
```
fred-apex-ai, line 43: It would be safeter to take a duration as argument rather than casting a double... but it's beyond porting to ros2, so up to you as I don't know where else that would require changes

control/velocity_controller/include/velocity_controller/velocity_controller.h
```diff
@@ -181,16 +177,14 @@ class VelocityController
   Lpf1d lpf_vel_error_;
 
   // methods
-  void callbackCurrentVelocity(const geometry_msgs::TwistStamped::ConstPtr & msg);
-  void callbackTrajectory(const autoware_planning_msgs::TrajectoryConstPtr & msg);
-  void callbackTimerControl(const ros::TimerEvent & event);
-  void callbackConfig(
-    const velocity_controller::VelocityControllerConfig & config, const uint32_t level);
+  void callbackCurrentVelocity(const geometry_msgs::msg::TwistStamped::ConstSharedPtr msg);
```
fred-apex-ai, line 180: I always had to change to `SharedPtr`. I wonder if/why that wouldn't be needed here. Are the callbacks not used?

control/velocity_controller/package.xml
```diff
@@ -1,23 +1,29 @@
...
   <depend>tf2_ros</depend>
-  <depend>geometry_msgs</depend>
   <depend>autoware_planning_msgs</depend>
-  <depend>autoware_control_msgs</depend>
-  <depend>dynamic_reconfigure</depend>
+  <depend>autoware_vehicle_msgs</depend>
 
-  <buildtool_depend>catkin</buildtool_depend>
+  <exec_depend>rclcpp</exec_depend>
+  <exec_depend>rclcpp_components</exec_depend>
```
fred-apex-ai, line 23: Do you really need the `*components`?

control/velocity_controller/package.xml
```diff
@@ -1,23 +1,29 @@
...
-
+  <version>0.0.1</version>
+  <description>vehicle velocity controller as a ROS 2 node</description>
   <maintainer email="makoto.kurihara@tier4.jp">Makoto Kurihara</maintainer>
+  <maintainer email="horibe.takamasa@gmail.com">Takamasa HORIBE</maintainer>
+  <license>Apache License 2.0</license>
 
-  <license>Appache2</license>
+  <buildtool_depend>ament_cmake</buildtool_depend>
+  <build_depend>rclcpp</build_depend>
```
fred-apex-ai, line 12: I think you should be a `<depend>`, else you miss one of the three depend properties for another package depending on this one

fred-apex-ai (changes requested):
Good job! Only minor changes required.

control/velocity_controller/cfg/VelocityController.cfg
```diff
@@ -1,69 +0,0 @@
-#! /usr/bin/env python
-
-PACKAGE = 'velocity_controller'
-
-from dynamic_reconfigure.parameter_generator_catkin import *
```
fred-apex-ai, line 5: I suppuose no callback are used so here you don't implement any dynamic reconfigure capabilities? Just checking because `velocity controller` sounds like you might want to tune in the car. I'm porting the MPC_follower and that needs

control/velocity_controller/include/velocity_controller/velocity_controller.h
TakaHoribe: I think we can use `ConstSharedPtr` here. There is also a subscriber using this callback here.
Which is the code you had to change to the `SharedPtr`?

control/velocity_controller/include/velocity_controller/delay_compensation.h
TakaHoribe: Done. Thanks!

control/velocity_controller/include/velocity_controller/delay_compensation.h
TakaHoribe: I had the same concern, but it should be done in the refactoring phase after porting.

control/velocity_controller/package.xml
TakaHoribe: Not needed. Thanks!

control/velocity_controller/package.xml
TakaHoribe: Thanks. Changed to <depend>.

control/velocity_controller/cfg/VelocityController.cfg
TakaHoribe: Sorry, I should add some comments on this point. I modified the top comment. Please check it.

control/velocity_controller/include/velocity_controller/delay_compensation.h
TakaHoribe: Sorry, I misunderstood your comment.

First, I tried to write like
```
pred_pose.header.stamp += rclcpp::Duration::from_seconds(delay_time);
```
but it couldn't. 

I just changed to the easy way, but do you know the better way to add the `double` duration to the `builtin_interface::Time`?

control/velocity_controller/CMakeLists.txt
TakaHoribe: Do you mean, to write like this?
```
ament_auto_add_executable(velocity_controller
  include/velocity_controller/delay_compensation.h
  include/velocity_controller/lowpass_filter.h
  include/velocity_controller/pid.h
  include/velocity_controller/velocity_controller_mathutils.h
  include/velocity_controller/velocity_controller.h
  src/velocity_controller_core.cpp
  src/velocity_controller_node.cpp
  src/pid.cpp
  src/velocity_controller_mathutils.cpp
)
```

In cmake, we usually do like
```
include_directories(include)
```
and, I'm not familiar with the ament, but I thought the ament automatically do include_directories like above, and the headers are not needed to add here.

To "help IDEs to discover which source file are part of the project" is the function of the ament? Or your IDE?

TakaHoribe:
@fred-apex-ai Thank you for your review! I replied and closed for some topics. Still some of them are open, and please check my comment.
Also, it would be nice if you can check the implementation of the `paramCallback()` as the porting of the `dynamic reconfigure` function.

control/velocity_controller/CMakeLists.txt
fred-apex-ai: > In cmake, we usually do like
> include_directories(include)

That's to inform the compiler  where to look for files and becomes `c++ -Iinclude/`

In contrast, what I meant is that adding the header files as sources actually tells cmake (and any IDE that queries cmake) that these files are part of this target `velocity_controller` and thus parf of this cmake project and not just files that happen to be in a folder. See e.g. this quote from https://crascit.com/2016/01/31/enhanced-source-file-handling-with-target_sources/

> In the above example, note that .h header files were specified as sources too, not just the .cpp implementation files. Headers listed as sources don’t get compiled directly on their own, but the effect of adding them is for the benefit of IDE generators like Visual Studio, Xcode, Qt Creator, etc. This causes those headers to be listed in the project’s file list within the IDE, even if no source file refers to it via #include. This can make those headers easier to find during development and potentially aid things like refactoring functionality, etc.

fred-apex-ai: > Do you mean, to write like this?

yes, that's what I mean

control/velocity_controller/include/velocity_controller/delay_compensation.h
fred-apex-ai: > I just changed to the easy way, but do you know the better way to add the double duration to the builtin_interface::Time?

I looked it up, there is no operator `+=` defined only `+`, the one you use http://docs.ros2.org/latest/api/rclcpp/namespacerclcpp.html#a31b2ba48a94966a93b36de2e52cae2bb

fred-apex-ai: > I had the same concern, but it should be done in the refactoring phase after porting.

I agree. we could add a TODO comment as simplest reminder

control/velocity_controller/include/velocity_controller/velocity_controller.h
fred-apex-ai: I think it was [here](https://github.com/tier4/Pilot.Auto/pull/7/files#diff-357aaf098422617a99a3b414b76d9b642d43edf3699d7cf4e2b67cbf0d2f0300R40) but now I see that it was a `const *`, not `ConstSharedPtr`. Never mind, keep your code as is

control/velocity_controller/CMakeLists.txt
TakaHoribe: Thank you! I modified the Cmakelist. Please check it.

control/velocity_controller/include/velocity_controller/delay_compensation.h
TakaHoribe: I added TODO comment.

fred-apex-ai (approved):
Good job Taka, I only have minor comments left. Will take this as an example to support parameter changes with callbacks

control/velocity_controller/src/velocity_controller_core.cpp
```diff
@@ -219,60 +234,109 @@ void VelocityController::callbackTimerControl(const ros::TimerEvent & event)
...
+  // PID limits
+  {
+    double max_pid = get_parameter("pid_controller/max_out").as_double();
+    double min_pid = get_parameter("pid_controller/min_out").as_double();
+    double max_p = get_parameter("pid_controller/max_p_effort").as_double();
+    double min_p = get_parameter("pid_controller/min_p_effort").as_double();
+    double max_i = get_parameter("pid_controller/max_i_effort").as_double();
+    double min_i = get_parameter("pid_controller/min_i_effort").as_double();
+    double max_d = get_parameter("pid_controller/max_d_effort").as_double();
+    double min_d = get_parameter("pid_controller/min_d_effort").as_double();
```
fred-apex-ai, line 324: That's a lot of code duplication. It would be much easier to put the two operations `get` and `set` into a function

control/velocity_controller/src/velocity_controller_core.cpp
```diff
@@ -376,13 +441,14 @@ CtrlCmd VelocityController::calcCtrlCmd()
   is_smooth_stop_ = checkSmoothStop(closest_idx, target_vel);
   if (is_smooth_stop_) {
     if (!start_time_smooth_stop_) {
-      start_time_smooth_stop_ = std::make_shared<ros::Time>(ros::Time::now());
+      start_time_smooth_stop_ = std::make_shared<rclcpp::Time>(get_clock()->now());
```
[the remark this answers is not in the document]
TakaHoribe: It is to determine if the timer is running by `if (!start_time_smooth_stop_)`.

control/velocity_controller/include/velocity_controller/velocity_controller.h
```diff
@@ -17,56 +17,57 @@
...
 #include <memory>
 #include <string>
 #include <vector>
 
-#include <geometry_msgs/PoseStamped.h>
-#include <geometry_msgs/TwistStamped.h>
-#include <ros/ros.h>
-#include <std_msgs/Bool.h>
-#include <std_msgs/Float32MultiArray.h>
+#include <geometry_msgs/msg/pose_stamped.hpp>
```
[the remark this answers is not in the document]
TakaHoribe: done

control/velocity_controller/src/velocity_controller_core.cpp
```diff
@@ -24,156 +24,170 @@ double lowpass_filter(const double current_value, const double prev_value, const
...
+  closest_dist_thr_ = declare_parameter("closest_waypoint_distance_threshold", 3.0);
+  closest_angle_thr_ = declare_parameter("closest_waypoint_angle_threshold", M_PI_4);
 
   // parameters for stop state
-  pnh_.param("stop_state_vel", stop_state_vel_, 0.0);                                // [m/s]
-  pnh_.param("stop_state_acc", stop_state_acc_, -2.0);                               // [m/s^2]
-  pnh_.param("stop_state_entry_ego_speed", stop_state_entry_ego_speed_, 0.2);        // [m/s]
-  pnh_.param("stop_state_entry_target_speed", stop_state_entry_target_speed_, 0.1);  // [m/s]
-  pnh_.param("stop_state_keep_stopping_dist", stop_state_keep_stopping_dist_, 0.5);  // [m/s]
+  stop_state_vel_ = declare_parameter("stop_state_vel", 0.0);                          // [m/s]
```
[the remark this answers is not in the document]
TakaHoribe: Yes, it's better to use like `velocity_mps`.

control/velocity_controller/src/velocity_controller_core.cpp
```diff
@@ -219,60 +234,109 @@ void VelocityController::callbackTimerControl(const ros::TimerEvent & event)
   resetHandling(control_mode_);
 }
 
-void VelocityController::callbackConfig(
-  const velocity_controller::VelocityControllerConfig & config, const uint32_t level)
+rcl_interfaces::msg::SetParametersResult VelocityController::paramCallback(
+  const std::vector<rclcpp::Parameter> & parameters)
 {
+  auto set_double_param_if_changed = [&](const std::string name, double & v) {
+    for (const auto & p : parameters) {
```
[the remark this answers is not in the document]
TakaHoribe: The `parameters` is not a vector of string, but a vector of `rclcpp::Parameter`. It needs to check `parameters[i].get_name() == name`, and no iterator for the parameter names. So I don't think we can use the `find` method here.

control/velocity_controller/src/velocity_controller_core.cpp
TakaHoribe: The setter function is already implemented in the original src code, and we need to modify them to combine with the new getter function defined here.

There should be a more effective way, e.g. having these parameters in the class member, and it should be done after porting. I'll write it as a TODO comment here.

control/velocity_controller/src/velocity_controller_core.cpp
```diff
@@ -219,60 +234,109 @@ void VelocityController::callbackTimerControl(const ros::TimerEvent & event)
   resetHandling(control_mode_);
 }
 
-void VelocityController::callbackConfig(
-  const velocity_controller::VelocityControllerConfig & config, const uint32_t level)
+rcl_interfaces::msg::SetParametersResult VelocityController::paramCallback(
+  const std::vector<rclcpp::Parameter> & parameters)
 {
+  auto set_double_param_if_changed = [&](const std::string name, double & v) {
```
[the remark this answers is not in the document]
TakaHoribe: changed to `update_param`. (and added `&`. Thanks!)

mitsudome-r:
@TakaHoribe Will it be ready to merge after adding this TODO comment?
https://github.com/tier4/Pilot.Auto/pull/27#discussion_r508923492

control/velocity_controller/src/velocity_controller_core.cpp
fred-apex-ai: alright, go ahead for now @TakaHoribe 
I'm having the same issue right now and found a different solution. Will ping you when the PR is ready :)

TakaHoribe:
@mitsudome-r Not yet. I also need to modify the `waitForTransform`. Working on it now.

control/velocity_controller/src/velocity_controller_core.cpp
```diff
@@ -24,156 +24,170 @@ double lowpass_filter(const double current_value, const double prev_value, const
...
-  while (ros::ok()) {
+  // set parameter callback
+  set_param_res_ =
+    this->add_on_set_parameters_callback(std::bind(&VelocityController::paramCallback, this, _1));
+
+  // wait to get vehicle position
+  while (rclcpp::ok()) {
     if (!updateCurrentPose(5.0)) {
-      ROS_INFO("waiting map to base_link at initialize.");
+      RCLCPP_WARN(get_logger(), "waiting map to base_link at initialize.");
```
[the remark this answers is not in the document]
TakaHoribe: I updated. Thank you!

TakaHoribe:
@fred-apex-ai I updated and replied to your reviews. Could you check them?

control/velocity_controller/src/velocity_controller_core.cpp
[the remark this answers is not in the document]
fred-apex-ai: I don't see any update, you are still calling `sleep_for`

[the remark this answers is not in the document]
fred-apex-ai: This is how I solved it with `waitForTransform` https://github.com/fred-apex-ai/autoware.iv.universe/commit/32f75580b6c6c445d2424261a7fc7726f3db3e75#diff-30dcb55e06b0d11249fb5d74414c4c96ceb4b891a18142afed00452ed9ef2ec3R148

control/velocity_controller/src/velocity_controller_core.cpp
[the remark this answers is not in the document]
fred-apex-ai: And of course `std::vector::find()` doesn't exist, one can use `std::find_if` as I did here https://github.com/fred-apex-ai/autoware.iv.universe/commit/32f75580b6c6c445d2424261a7fc7726f3db3e75#diff-30dcb55e06b0d11249fb5d74414c4c96ceb4b891a18142afed00452ed9ef2ec3R178

fred-apex-ai (changes requested):

TakaHoribe:
@fred-apex-ai I have an exception issue.

When I run the velocity_controller, I've got this exception. It occurs only once after the initial pose is published.

```
[velocity_controller-1] terminate called after throwing an instance of 'tf2::LookupException'
[velocity_controller-1]   what():  Failed to transform from base_link to map
```


How to reproduce this is below.
```
$ ros2 launch velocity_controller velocity_controller.launch.xml

$ ros2 launch simple_planning_simulator simple_planning_simulator.launch.xml

$ rviz2

Then, publish initialpose in the Rviz.
```

fred-apex-ai:
@nnmm Do you have an idea? Is it perhaps because tf first needs to create the integral tree representation, so a lookup only works on the 2nd message?

nnmm:
The `waitForTransform()` throws that exception? No, I have never had that happen. I can take a closer look later.

fred-apex-ai:
@TakaHoribe I took over your approach into https://github.com/tier4/Pilot.Auto/pull/52 and didn't get an exception on the first lookup. Should I continue to investigate or did you resolve the issue in the meantime?

TakaHoribe:
@fred-apex-ai Sorry, I couldn’t have an enough time to investigate it today. Your implementation is almost same as mine, and if it doesn’t give the exception, the problem is maybe in a different place than WaitForTransform() in the velocity_controller.

Ill take this issue. Thank you for your testing.

TakaHoribe:
@fred-apex-ai I tested with the mpc_follower, and the error didn't occur. One difference is you don't use the returned value from the `waitForTransform()`.
In my code, I get the transform like this.

```
auto tf_future = tf_buffer_.waitForTransform(input, output, tf2::TimePointZero, 
    tf2::durationFromSec(0.0), [](auto &) {});
const auto status = tf_future.wait_for(timeout);
if (status == std::future_status::ready) {
      transform = tf_future.get();  // this causes the  tf2::LookupException
}
```

In your code,
```
// in the blockUntilVehiclePositionAvailable()

auto tf_future = tf_buffer_.waitForTransform(input, output, tf2::TimePointZero, 
    tf2::durationFromSec(0.0), [](auto &) {});
const auto status = tf_future.wait_for(timeout);
if (status == std::future_status::ready) {
     return;
}

// Then updateCurrentPose(),

transform = tf_buffer_.lookupTransform("map", 
    "base_link", rclcpp::Time(0));  // this doesn't causes any exception.
```

Your code works correctly, so I just change to use your way. Still don't know why tf_future throws exception even the tf is available.

I also tested adding the `tf_future.get()` [here](https://github.com/tier4/Pilot.Auto/blob/81b32a777bc7400160cfb6eec5343acc35c4aa3d/control/mpc_follower/src/mpc_follower_core.cpp#L1003) in the mpc_follower, then it caused the same exception.

TakaHoribe:
@fred-apex-ai I have applied your method. Could you review this again?
The functions `blockUntilVehiclePositionAvailable()` and `updateCurrentPose()` have been rewritten. It's easier to see the actual code than the difference.

control/velocity_controller/src/velocity_controller_core.cpp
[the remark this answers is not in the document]
TakaHoribe: Changed to the same way of `mpc_follower`.

TakaHoribe:
@fred-apex-ai I keep the `blockUntilVehiclePositionAvailable` function since `canTransform` and `lookupTransform` gives a warning with high frequency until the `map`->`base_link` tf is available. See https://github.com/tier4/Pilot.Auto/pull/27/commits/ef61c2a77e43bd86afb2ac73f50ae14c57555c48 . (Also changed the timeout sec from 5s to 1s since it is just a sleep duration, not timeout.)

Except for this, almost the same as the mpc. After your look, I'll merge this PR.

fred-apex-ai:
@TakaHoribe Ok, it's fine to me. I noticed that message in the terminal as well. THe MPC follower will emit that, so our solutions now aren't 100 % identical but that's ok. Ready for merge finally! It's been a long ride with this PR

TakaHoribe:
@fred-apex-ai Thank you for your cooperation for a long time!

PR #28, opened 2020-10-14
Rename launch files to launch.xml

nnmm:
This should help GitHub with showing the diff side-by-side in the porting PRs instead of treating them as separate files.


mitsudome-r:
@nnmm Thanks, I think this is a good idea.
Could you fix the conflict? I will make sure that this will be merged first before any other open PRs.

PR #29, opened 2020-10-14
Port ndt to ROS2

nnmm:
Very similar to https://github.com/tier4/Pilot.Auto/pull/20 and https://github.com/tier4/Pilot.Auto/pull/21, see these PRs for discussion. CI is still failing as long as those are not merged.


localization/pose_estimator/ndt_scan_matcher/ndt/CMakeLists.txt
```diff
@@ -1,21 +1,17 @@
...
-
-catkin_package(
-  INCLUDE_DIRS include
-)
-
-include_directories(include ${catkin_INCLUDE_DIRS} ${PCL_INCLUDE_DIRS})
-link_directories(${PCL_LIBRARY_DIRS})
-add_definitions(${PCL_DEFINITIONS})
+find_package(ament_cmake REQUIRED)
+find_package(PCL REQUIRED COMPONENTS common io registration)
```
nnmm, line 12: Interestingly, I got an error when PCL came after ndt_omp and ndt_pcl_modified:
```

CMake Error at CMakeLists.txt:16 (add_library):
  Target "ndt" links to target "ndt_omp::align" but the target was not found.
  Perhaps a find_package() call is missing for an IMPORTED target, or an
  ALIAS target is missing?


CMake Generate step failed.  Build files cannot be regenerated correctly.
```
Not sure why.

esteve:
@nnmm could you migrate this PR to use ament_auto? It should be fairly straightforward. Also, now that the issue with DEBIAN_FRONTEND and sudo is fixed, could you rebase this on top of ros2? Thanks.

nnmm:
@esteve ament_auto didn't play well with PCL in my other PRs: it doesn't export the build dependency on PCL so that downstream builds break. In this case, the only downstream package happens to pull in PCL headers some other way, so you could leave it, but it's brittle.

Plus I suspect that only the manual style allows choosing only a subset of PCL components, which could potentially give smaller/faster builds.

nnmm:
@esteve Do you have any more comments? Otherwise, if you can approve before Tuesday evening, I'd then rebase onto the newest ros2 branch such that CI can pass and then only need to find someone to merge it. That would unblock the next PR (ndt_scan_matcher) in this dependency chain.

esteve (approved):
LGTM

PR #33, opened 2020-10-16
Port autoware_utils

nnmm:
This package is not in the list of packages to port. However, perhaps this is a good place for time conversion functions, which I need for ndt_scan_matcher.

I changed nothing interesting, I only added one missing header.


mitsudome-r (approved):
LGTM

PR #38, opened 2020-10-19
port astar_search to ros2

TakaHoribe:
needed by https://github.com/tier4/Pilot.Auto/pull/41.

 - Since `tf2::doTransform(Pose, Pose, Transform)` is not in the ROS2 tf_geometry_msgs, it is replaced to `tf2::doTransform(PoseStamped, PoseStamped, Transform)`

### DIFF
--- a/planning/scenario_planning/parking/astar_search/CMakeLists.txt
+++ b/planning/scenario_planning/parking/astar_search/CMakeLists.txt
@@ -1,48 +1,18 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(astar_search)
 
-add_compile_options(-std=c++14)
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
 
-find_package(catkin REQUIRED COMPONENTS
-  geometry_msgs
-  nav_msgs
-  std_msgs
-  tf2
-  tf2_geometry_msgs
-)
-
-catkin_package(
-  INCLUDE_DIRS include
-  LIBRARIES astar_search
-)
-
-include_directories(
-  include
-  ${catkin_INCLUDE_DIRS}
-)
-
-add_library(astar_search
-  src/astar_search.cpp
-)
-
-target_link_libraries(astar_search
-  ${catkin_LIBRARIES}
-)
-
-add_dependencies(astar_search
-  ${catkin_EXPORTED_TARGETS}
-)
-
-install(DIRECTORY include/${PROJECT_NAME}/
-  DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-)
-
-install(TARGETS astar_search
-  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-)
-
-if(CATKIN_ENABLE_TESTING)
-  find_package(rostest REQUIRED)
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
 endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wno-unused-parameter -Wall -Wextra -Wpedantic)
+endif()
+
+ament_auto_add_library(astar_search src/astar_search.cpp)
+
+ament_auto_package(
+  INSTALL_TO_SHARE
+)
--- a/planning/scenario_planning/parking/astar_search/CMakeLists.txt
+++ b/planning/scenario_planning/parking/astar_search/CMakeLists.txt
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 3.5)
 project(astar_search)
 
-find_package(ament_cmake_auto REQUIRED)
-ament_auto_find_build_dependencies()
-
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wno-unused-parameter -Wall -Wextra -Wpedantic)
 endif()
+
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
 
 ament_auto_add_library(astar_search src/astar_search.cpp)
 
--- a/planning/scenario_planning/parking/astar_search/include/astar_search/astar_search.h
+++ b/planning/scenario_planning/parking/astar_search/include/astar_search/astar_search.h
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-#ifndef ASTER_PLANNER_H
-#define ASTER_PLANNER_H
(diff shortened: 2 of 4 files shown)


planning/scenario_planning/parking/astar_search/include/astar_search/astar_search.h
```diff
@@ -17,17 +17,18 @@
 #ifndef ASTER_PLANNER_H
```
jilaada, line 17: [Minor]: There is a typo here - not sure if its important for this PR though.

planning/scenario_planning/parking/astar_search/include/astar_search/astar_search.h
```diff
@@ -17,17 +17,18 @@
...
 #define ASTER_PLANNER_H
 
+#include <cmath>
 #include <functional>
 #include <iostream>
 #include <queue>
 #include <string>
 #include <tuple>
 #include <vector>
```
jilaada, line 26: [Minor]: Since I can only comment on it on this file, I don't know if the vector include is needed (in `astar_search.cpp`) since `vector` is already included in the `astar_search.h`?

jilaada (commented):
I was able to compile this package in the ADE Foxy environment, though unable to test it. I have some minor comment which are mostly cosmetic. Otherwise this LGTM. :+1:

planning/scenario_planning/parking/astar_search/CMakeLists.txt
```diff
@@ -1,48 +1,18 @@
...
-)
-
-target_link_libraries(astar_search
-  ${catkin_LIBRARIES}
-)
-
-add_dependencies(astar_search
-  ${catkin_EXPORTED_TARGETS}
-)
+if(NOT CMAKE_CXX_STANDARD)
```
jilaada, line 7: [Minor]: Not sure if order particularly matters in this case but it seems like most poeple are specifying the compile options before the rest of the build process. Not sure if that should be strictly followed - everything compile so maybe its not an issue.

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
```diff
@@ -195,18 +199,18 @@ void AstarSearch::initializeNodes(const nav_msgs::OccupancyGrid & costmap)
   // Initialize nodes
   nodes_.clear();
   nodes_.resize(height);
-  for (int i = 0; i < height; i++) {
+  for (uint32_t i = 0; i < height; i++) {
```
jilaada, line 202: Maybe I'm not understanding the change but since these are used for indexing, should these be `size_t` instead?

planning/scenario_planning/parking/astar_search/CMakeLists.txt
TakaHoribe: This should be the top of this file. Thanks!

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
TakaHoribe: There were some warnings for the comparison between `(int) i` and `(unsigned int) height`. Yes, `size_t` is also fine.

planning/scenario_planning/parking/astar_search/include/astar_search/astar_search.h
TakaHoribe: It is not needed. I removed it.

planning/scenario_planning/parking/astar_search/include/astar_search/astar_search.h
TakaHoribe: I fixed. Also added `SCENARIO_PLANNING_` prefix.

planning/scenario_planning/parking/astar_search/package.xml
```diff
@@ -1,17 +1,22 @@
...
+<package format="3">
   <name>astar_search</name>
   <version>0.1.0</version>
   <description>The astar_search package</description>
   <maintainer email="aohsato@gmail.com">Akihito Ohsato</maintainer>
   <author email="aohsato@gmail.com">Akihito Ohsato</author>
-  <license>BSD</license>
+  <license>Apache License 2.0</license>
 
-  <buildtool_depend>catkin</buildtool_depend>
```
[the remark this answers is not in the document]
TakaHoribe: My mistake. Thanks!

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
```diff
@@ -346,8 +350,8 @@ bool AstarSearch::search()
 
 void AstarSearch::setPath(const AstarNode & goal_node)
 {
-  std_msgs::Header header;
-  header.stamp = ros::Time::now();
+  std_msgs::msg::Header header;
+  header.stamp = rclcpp::Clock().now();
```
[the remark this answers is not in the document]
TakaHoribe: This is not a rosnode, so `this->get_clock()` is not an option. I'm checking the difference between `rclcpp::Node::now()` and `rclcpp::Clock().now()`.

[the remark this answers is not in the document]
TakaHoribe: @jilaada In this case (no Nodes are defined), I cannot see any differences between `rclcpp::Node::now()` and `rclcpp::Clock().now()`. How do you think?

http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Node.html#a1db9cec1af19a994b3f1dad2059d2707

TakaHoribe:
@jilaada I replied all of your comments. Please check them.

> though unable to test it

Maybe you can test with freespace_planner https://github.com/tier4/Pilot.Auto/pull/41 , which uses this astar_search library in the executable.

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
```diff
@@ -14,56 +14,58 @@
  * limitations under the License.
  */
 
-#include "astar_search/astar_search.h"
-
-#include <vector>
+#include <astar_search/astar_search.h>
```
jilaada, line 17: [Minor]: I think the original usage of `""` instead of `<>` is correct when referencing files local to this project?

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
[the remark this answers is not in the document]
jilaada: @TakaHoribe Here is some reference issue I found regarding the difference between Node and Clock time: https://github.com/ros2/rclcpp/issues/595. It seems like Node time allows for the use of sim time when use sim time is enabled. It seems like Clock only takes the the system time. But since this is not a node, I'm not sure what the behaviour will be? :thinking:

jilaada (commented):
Thanks for the fixes, a couple more questions from me though.

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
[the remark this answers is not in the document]
mitsudome-r: I never checked but I thought it should be able to refer sim time if you give RCL_ROS_TIME as constructor.
http://docs.ros2.org/dashing/api/rcl/time_8h.html#a5c734f508ce06aec7974af10ad09d071

i.e. 
rclcpp::Clock(RCL_ROS_TIME).now();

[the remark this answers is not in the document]
jilaada: Yeah this would be the best option I think. The default clock implementation uses the `RCL_SYSTEM_TIME` which would not be simulator friendly: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/clock.hpp#L66

TakaHoribe:
@jilaada Sorry for the late response. I fixed two points you mentioned (include format and clock). Please check them.

jilaada (approved):
Changes look good to me! :+1:

PR #39, opened 2020-10-19
ROS2 Porting: gyro_odometer

jilaada:
# gyro_odometer

## Summary

Port `gyro_odometer` package. This package was straightforward with not many dependencies.

## Testing

```
colcon build
. install/setup.bash
```
2. Launch the package
```
ros2 launch gyro_odometer gyro_odometer.launch.xml
```

### DIFF
--- a/localization/twist_estimator/gyro_odometer/include/gyro_odometer/gyro_odometer_core.h
+++ b/localization/twist_estimator/gyro_odometer/include/gyro_odometer/gyro_odometer_core.h
@@ -41,16 +41,13 @@ private:
     const std::string & target_frame, const std::string & source_frame,
     const geometry_msgs::msg::TransformStamped::SharedPtr transform_stamped_ptr);
 
-  // ros::NodeHandle nh_;
-  // ros::NodeHandle private_nh_;
+  rclcpp::Subscription<geometry_msgs::msg::TwistStamped>::SharedPtr vehicle_twist_sub_;
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub_;
 
-  // ros::Subscriber vehicle_twist_sub_;
-  // ros::Subscriber imu_sub_;
-
-  // ros::Publisher twist_pub_;
-  // ros::Publisher twist_with_covariance_pub_;
-  // ros::Publisher linear_x_pub_;
-  // ros::Publisher angular_z_pub_;
+  rclcpp::Publisher<geometry_msgs::msg::TwistStamped>::SharedPtr twist_pub_;
+  rclcpp::Publisher<geometry_msgs::msg::TwistWithCovarianceStamped>::SharedPtr twist_with_covariance_pub_;
+  // rclcpp::Publisher<std_msgs::Float32>::SharedPtr linear_x_pub_;
+  // rclcpp::Publisher<std_msgs::Float32>::SharedPtr angular_z_pub_;
 
   std::shared_ptr<tf2_ros::Buffer> tf_buffer_;
   std::shared_ptr<tf2_ros::TransformListener> tf_listener_;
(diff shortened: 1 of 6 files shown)


localization/twist_estimator/gyro_odometer/include/gyro_odometer/gyro_odometer_core.h
```diff
@@ -14,47 +14,46 @@
...
-  ros::NodeHandle nh_;
-  ros::NodeHandle private_nh_;
+  rclcpp::Subscription<geometry_msgs::msg::TwistStamped>::SharedPtr vehicle_twist_sub_;
+  rclcpp::Subscription<sensor_msgs::msg::Imu>::SharedPtr imu_sub_;
 
-  ros::Subscriber vehicle_twist_sub_;
-  ros::Subscriber imu_sub_;
+  rclcpp::Publisher<geometry_msgs::msg::TwistStamped>::SharedPtr twist_pub_;
+  rclcpp::Publisher<geometry_msgs::msg::TwistWithCovarianceStamped>::SharedPtr twist_with_covariance_pub_;
+  // rclcpp::Publisher<std_msgs::Float32>::SharedPtr linear_x_pub_;
```
nnmm, line 49: Are the commented out publishers not needed anymore?

localization/twist_estimator/gyro_odometer/launch/gyro_odometer.launch.xml
```diff
@@ -1,5 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nnmm, line 1: Should we do this for other packages as well? I haven't seen it yet.

localization/twist_estimator/gyro_odometer/src/gyro_odometer_core.cpp
```diff
@@ -16,49 +16,52 @@
...
-: nh_(nh), private_nh_(private_nh), output_frame_("base_link"), tf2_listener_(tf2_buffer_)
+#include <cmath>
+
+GyroOdometer::GyroOdometer()
+: Node("map_tf_generator"), 
+  output_frame_(declare_parameter("base_link", "base_link"))
 {
-  private_nh_.getParam("output_frame", output_frame_);
+  rclcpp::Clock::SharedPtr clock = std::make_shared<rclcpp::Clock>(RCL_ROS_TIME);
+  tf_buffer_ = std::make_shared<tf2_ros::Buffer>(clock);
```
nnmm, line 28: We could also make them plain objects instead of shared pointers. I find it nice to use plain objects by default, what do you think?

nnmm (commented):
LGTM as soon as I understand what's going on with the commented out publishers :+1:

localization/twist_estimator/gyro_odometer/include/gyro_odometer/gyro_odometer_core.h
jilaada: Originally, these were not commented out however in the constructor, the publisher were never created (commented out) in the original implementation. I ported these regardless in case they are needed

localization/twist_estimator/gyro_odometer/launch/gyro_odometer.launch.xml
jilaada: I don't think it makes much of a difference, I can remove it if it causes confusion

localization/twist_estimator/gyro_odometer/src/gyro_odometer_core.cpp
jilaada: Sure, I can do that. I was using smart pointers as this was how it was done in the rclcpp test code for these objects - though that's not really an excuse to use them. I can change this to use plain objects

localization/twist_estimator/gyro_odometer/include/gyro_odometer/gyro_odometer_core.h
nnmm: Ah, makes sense.

nnmm (approved):

localization/twist_estimator/gyro_odometer/src/gyro_odometer_core.cpp
nnmm: I'll leave it up to you.

PR #40, opened 2020-10-20
ROS2 Porting: emergency_handler

jilaada:
# emergency_handler

## Summary

Port the emergency_handler package to ROS2. Simple package with not many dependencies so the conversion was straightforward. Key changes:
* Convert headers and type declarations to ROS2
* Convert logging 
* Use of timers
* Conversion of parameter configuration files

## Testing

```
colcon build
. install/setup.bash
```
2. Launch the package
```
ros2 launch emergency_handler emergency_handler.launch.xml
```
3. Check to make sure that the configuration file parameters have been loaded correctly
```
ros2 param list
```

## Discussion Points

1. Sorry about the annoying diff view for some of the files, this package had a strange structure to it and I wanted to change, as a result some of the packages have been moved and renamed. If this is an unnecessary change I can revert these easily.


nnmm (approved):
Fred and I load balanced PRs a bit between us, so I reviewed – this looks good to me!

system/emergency_handler/config/emergency_handler.param.yaml
```diff
@@ -0,0 +1,6 @@
+# Default configuration for emergency handler
+---
+/emergency_handler:
```
mitsudome-r, line 3: Let's use `/**` instead of specifying node name so that parameter still applies even if the node name is changed in a launch file.

mitsudome-r (changes requested):

system/emergency_handler/config/emergency_handler.param.yaml
jilaada: Ah yeah good catch. I hadn't thought of that.

jilaada:
@mitsudome-r Seems like the repo doesn't restrict merging if a change is requested. I re-requested your review anyway

mitsudome-r (approved):
LGTM

PR #42, opened 2020-10-20
ROS2 Porting: autoware_error_monitor

jilaada:
# autoware_error_monitor

## Summary

Port the `autoware_error_monitor` package. Package itself is fairly straightforward, though it contains a lot of configuration files and launches the diagnostic_aggregator node. Notable changes:
* Modify parameter file
* Changing the names of the keys in the KeyName struct - this was used to access the parameters though in ROS1 nested parameters were delimited with a `/` while in ROS2 this has changed to `.`. [[ref](https://answers.ros.org/question/325939/declare-nested-parameter/)]

## Testing

```
colcon build
. install/setup.bash
```
2. Launch the package
```
# Launches the node ONLY
ros2 launch autoware_error_monitor autoware_error_monitor_node.launch.xml

# Launches the node along with the diagnostic_aggregator
ros2 launch autoware_error_monitor autoware_error_monitor.launch.xml

# Launches dummy diag publishes (not ported yet in ROS2)
ros2 launch autoware_error_monitor debug_tools.launch.xml 
```

## Discussion Points

Some unmet dependencies:
* [x] One of the launch file launches the `dummy_diag_publisher` node which hasn't been ported yet
* [x] Rebase to ros2
* [x] Fix possible issues with `diagnostic_aggregator` configuration files


system/autoware_error_monitor/config/diagnostic_aggregator/_discard.yaml
```diff
@@ -1,9 +1,15 @@
...
-    path: speaker
-    contains: ": speaker"
-  joystick:
-    type: diagnostic_aggregator/DiscardAnalyzer
-    path: joystick
-    contains: ": Joystick"
+/**:
+  ros__parameters:
+    path: system
+    discard:
```
jilaada, line 4: I needed to change the layout here in order for the `DiscardAnalyzers` to be picked up else these would be entirely skipped for some reason by the Diagnostics Aggregator.

jilaada:
I have done some testing with the `diagnostics_aggregator`. To test I used the `rqt_robot_monitor` which was available in foxy just needs to be installed via apt or rosdep (your preference - I have not added this as an `exec_depend` in the `package.xml` as it is not run via the launch file. Here is a capture of the output:

Screenshot from 2020-11-05 17-04-18

I've never used the diagnostics_aggregator and I'm not sure what to be expected. Running this node shows all `OK` statuses which would imply nominal working behaviour but nothing is running... so really these should all be errors? Maybe I have misunderstood the purpose of this tool.

Here are some of the changes I needed to make and the caveats I found when trying to port the configuration files:

* `contains` field is now a string array
* Following the examples in the [rclcpp aggregator test code](https://github.com/ros/diagnostics/tree/foxy/diagnostic_aggregator/test), the type of aggregator needed to be put in quotations - though this seemed to work without anyway
* Requirement of a **single** root rather than multiple roots which with their own children. Interestingly when trying to use a different root path, the aggregator seems to just ignore this and groups all of the topics under any existing root. As a result all the config files now have the same root `system`
* DiscardAnalyzers are not showing up on `rqt_robot_monitor` but debug logs show that the `DiscardAnalyzer` has been created.

kenji-miyake:
@jilaada 

> so really these should all be errors? Maybe I have misunderstood the purpose of this tool.

Since it specifies `num_items: 0`, diagnostic_aggregator will output `OK` if there is no diagnostics that match `contains` condition.
https://github.com/ros/diagnostics/blob/322ba7071ab7368e0680e62d05aa63dfaf988ef6/diagnostic_aggregator/include/diagnostic_aggregator/generic_analyzer_base.hpp#L235-L238

If you publish some diagnostics (for example using dummy_diag_publisher), you can see them in rqt_robot_monitor.
I recommend you to try ROS1 version of this node to confirm the behavior.

Please let me know if you need more detail.

jilaada:
> num_items

Ah, thanks for the detailed reply! I was actually testing with the ROS1 via the `ros1_bridge` and interestingly it works. However it did not come up on `rqt_runtime_monitor` only `rqt_robot_monitor`.

kenji-miyake:
@jilaada Actually it seems your robot_monitor isn't working as expected. robot _monitor shows some data when the config is loaded even if there is no diag topics.

Could you tell me the results of `ros2 topic echo /diagnostics`. runtime_monitor is just visualizing this topic.

jilaada:
> @jilaada Actually it seems your robot_monitor isn't working as expected. robot _monitor shows some data when the config is loaded even if there is no diag topics.
> 
> Could you tell me the results of `ros2 topic echo /diagnostics`. runtime_monitor is just visualizing this topic.

[ROS1] Yeah so the `robot_monitor` does work but the `runtime_monitor` does not work as nothing is publishing on the `/diagnostics` topic

[ROS2] `runtime_monitor` is not available only `robot_monitor` which is what is displayed in the screenshot above. The diagnostics topic does not display anything - presumably because nothing is publishing diagnostics so this makes sense. I haven't run this with the `dummy_diag_publisher` yet though :thinking:

system/autoware_error_monitor/src/autoware_error_monitor_core.cpp
```diff
@@ -13,24 +13,59 @@
...
+    std::bind(&AutowareErrorMonitor::onDiagArray, this, std::placeholders::_1));
+
+  // Publisher
+  pub_driving_capability_ = create_publisher<autoware_system_msgs::msg::DrivingCapability>(
+    "output/driving_capability", rclcpp::QoS{1});
+
+  // Timer
+  auto timer_callback = std::bind(&AutowareErrorMonitor::onTimer, this);
+  auto period = std::chrono::duration_cast<std::chrono::nanoseconds>(
+    std::chrono::duration<double>(update_rate_));
```
mitsudome-r, line 45: You have to take an inverse of rate in order to get period.
```suggestion
    std::chrono::duration<double>(1.0 / update_rate_));
```

system/autoware_error_monitor/config/autoware_error_monitor.param.yaml
```diff
@@ -0,0 +1,26 @@
...
+      remote_control: ['control',
+        'remote_control',
+        'safety',
+        'vehicle'
+      ]
+      safe_stop: ['map',
+        'emergency_system',
+        'vehicle'
+      ]
+      emergency_stop: ['vehicle']
```
mitsudome-r, line 26: Since we set root path for /diagnostic_agg to `autoware` (currently `system`), these key should also be `autoware/vehicle` instead of `vehicle`. Same with other keys.

mitsudome-r:
I've tried it with dummy_diagnostic_publisher.
I had trouble making DiagnosticAggregator work properly, but I ended up finding out that it's due to queue size for subscription being too small.
I created an issue to ask the maintainer how we should fix this issue: https://github.com/ros/diagnostics/issues/167.

system/autoware_error_monitor/src/autoware_error_monitor_core.cpp
jilaada: Thanks, I missed the conversion on this one!

system/autoware_error_monitor/config/autoware_error_monitor.param.yaml
jilaada: Ok I wasn't aware of this, I will fix this up. Thanks for the information. Done in d38ebef

jilaada: I don't quite understand, sorry. When do we set the root root path of /diagnostics_agg to `autoware`? Do you mean that in the other configuration files for the diagnostic aggregator the `system` namespace should be changed to `autoware`? Or do you mean in this configuration file, the keys `vehicle`, `map`, `control`, etc should be prefixed with the `autoware/`?

mitsudome-r: since we changed the root of diagnostic_aggregator to `autoware` we also have to modify the keys that we specify in his param file accordingly.

jilaada: Done in 0e7c769

jilaada:
@mitsudome-r Any more comments here?

mitsudome-r:
Let's add diagnostics to repos file until  https://github.com/tier4/Pilot.Auto/pull/42#issuecomment-723119725 is merged into the upstream and released as foxy binary package. 
https://github.com/tier4/AutowareArchitectureProposal/pull/124

jilaada:
@mitsudome-r tier4/AutowareArchitectureProposal#124 is merged. I will go ahead and merge this PR too.

PR #48, opened 2020-10-20
Message definition for ndt_scan_matcher debug topics

nnmm:
* There are a [couple of `Float32` publishers](https://github.com/tier4/Pilot.Auto/blob/0f3dee10dd4c22fddbee44662bd520e5a6d87ef7/localization/pose_estimator/ndt_scan_matcher/ndt_scan_matcher/src/ndt_scan_matcher_core.cpp#L114-L122) in `ndt_scan_matcher` that are apparently for debugging; there are no corresponding subscribers in the code base.
* They all publish at the same time, so they can be made into one message instead of individual messages. This package defines that message type.
* When I compiled this with `ament_cmake_auto`, the message wasn't found by the `ndt_scan_matcher` package, so I used `ament_cmake` as in the other message packages.

### DIFF
--- a/localization/pose_estimator/ndt_scan_matcher/ndt_scan_matcher_msgs/CMakeLists.txt
+++ b/localization/pose_estimator/ndt_scan_matcher/ndt_scan_matcher_msgs/CMakeLists.txt
@@ -1,0 +1,19 @@
+cmake_minimum_required(VERSION 3.5)
+project(ndt_scan_matcher_msgs)
+
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+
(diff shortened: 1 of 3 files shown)


TakaHoribe:
@nnmm Thank you! But I have a concern to have a specific debug message type.

Debug message types are expected to change easily depending on the environments or projects in which they are used.
Because ROS is not flexible with message type changes, it is easier for users to use all debug topics as separate types.

If these topics are needed for architecture, they should be combined into one. But in this case, I think it's better to keep them separate.

However, std_msg is deprecated, so we will replace them with example_msgs or create the autoware_util_msg to publish Float.

nnmm:
Closing this issue unless the consensus re:https://github.com/tier4/AutowareArchitectureProposal/issues/92 changes.

PR #51, opened 2020-10-21
Port awapi_awiv_adapter to ROS2

nnmm:
Relatively straightforward port.

### DIFF
--- a/awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
+++ b/awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
@@ -21,13 +21,15 @@
 
 namespace autoware_api
 {
-class AutowareIvAutowareStatePublisher : public rclcpp::Node
+class AutowareIvAutowareStatePublisher
 {
 public:
-  AutowareIvAutowareStatePublisher();
+  AutowareIvAutowareStatePublisher(rclcpp::Node& node);
   void statePublisher(const AutowareInfo & aw_info);
 
 private:
+  rclcpp::Logger logger_;
+  rclcpp::Clock::SharedPtr clock_;
   // publisher
   rclcpp::Publisher<autoware_api_msgs::msg::AwapiAutowareStatus>::SharedPtr pub_state_;
 
--- a/awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
+++ b/awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
@@ -14,63 +14,60 @@
  * limitations under the License.
  */
 
-#include <autoware_api_msgs/AwapiAutowareStatus.h>
+#include <rclcpp/rclcpp.hpp>
+
 #include <awapi_awiv_adapter/awapi_autoware_util.h>
+#include <autoware_api_msgs/msg/awapi_autoware_status.hpp>
 
 namespace autoware_api
 {
-class AutowareIvAutowareStatePublisher
+class AutowareIvAutowareStatePublisher : public rclcpp::Node
 {
 public:
   AutowareIvAutowareStatePublisher();
   void statePublisher(const AutowareInfo & aw_info);
 
 private:
-  // node handle
-  ros::NodeHandle nh_;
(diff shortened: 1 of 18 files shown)


nnmm:
@mitsudome-r I will still take care of aligning the constness and reference-ness of parameters closer with the original code.

awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
```diff
@@ -14,63 +14,60 @@
...
-#include <autoware_api_msgs/AwapiAutowareStatus.h>
+#include <rclcpp/rclcpp.hpp>
+
 #include <awapi_awiv_adapter/awapi_autoware_util.h>
+#include <autoware_api_msgs/msg/awapi_autoware_status.hpp>
 
 namespace autoware_api
 {
-class AutowareIvAutowareStatePublisher
+class AutowareIvAutowareStatePublisher : public rclcpp::Node
```
mitsudome-r, line 24: Only `AutowareIvAdapter` class should be inheriting the `rclcpp::Node` class.
Otherwise, it will create another node.
(Same goes for `AutowareIvVehicleStatePublisher`, `AutowareIvStopReasonAggregator`, `AutowareIvLaneChangeStatePublisher`, AutowareIvObstacleAvoidanceStatePublisher`)

If you need access to the loggers or create pub/sub within this class, then we should pass `AutowareIvAdapter`'s loggers and topic_interface  (e.g. by `get_node_topics_interface ()`) or just pass `AutowareIvAdapter`'s node pointer to each class.

mitsudome-r (commented):
Good work on porting this large package.
I just have few comments.
Also, it looks like the launch file is not ported.

awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
nnmm: No problem, I can do that. I assumed it would be possible to create nested nodes, since the ROS1 code also had several `NodeHandles`, but I tested with a toy example and it doesn't work. I'm asking an expert in Apex.AI if there's a way to make it work, just for completeness.

nnmm:
@mitsudome-r You're right about the launch file, my editor for some reason failed to show it to me. What should we do with all the `topic_tools` relay nodes?

mitsudome-r:
to be honest, I can't think of a good way to solve topic relay issue. We might have to create our own relay node using generic_subscription and generic_publisher from rosbag2. https://github.com/ros2/rosbag2/tree/c34601acba77d08c22211d3eb062b8ba867fef71/rosbag2_transport/src/rosbag2_transport

awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
nnmm: Just making sure: I was informed that having nested nodes is possible if you add them to the executor, but you'd prefer there to be only one node, right?

nnmm:
Yes, they could be copied, since they have very few dependencies. We could also write our own node that has a subscription-publisher pair for all of the topics listed in the launch file.

nnmm:
Hmm, the two files have few dependencies but it looks like they're hard to instantiate without pulling in many additional libraries. I'm investigating if there exists an easy way to get the required `rosidl_message_type_support_t`.

awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.h
mitsudome-r: Since original ROS1 code is in one node, I think we want to keep it as one during porting.

nnmm:
This is now ready for the (perhaps) last round of reviewing, especially the launch file – I have simply changed the global `use_emergency_handling` parameter to a normal ROS2 parameter with default `false`, but I'm not sure that is correct.

I also used child loggers for the secondary classes, meaning logger names like "awapi_awiv_adapter_node.awapi_awiv_lane_change_state_publisher".

mitsudome-r (approved):

PR #58, opened 2020-10-26
vehicle_info_util package for ROS2 port

nnmm:
As discussed, I made this node into a simple utility class that fetches parameters and calculates derived parameters locally.

I'd like to add comments on what the "offset" variables mean – are those the offsets from a reference point on the ground below the middle of the rear axle?


mitsudome-r:
> I'd like to add comments on what the "offset" variables mean – are those the offsets from a reference point on the ground below the middle of the rear axle?

yes

nnmm:
I updated the comments, and also mentioned that the overhang values etc. describe the bounding box.

For an example of how usage of this struct looks, see [this draft](https://github.com/tier4/Pilot.Auto/pull/64/files#diff-802ee24659db98121c936879be0e12cf2e36b95e595239a5f98ccea2cae63d17). I also opened an issue, https://github.com/tier4/AutowareArchitectureProposal/issues/99, to track places that would need to be adapted in the same fashion and for general discussion.

vehicle/util/vehicle_info/src/vehicle_info.cpp
```diff
@@ -0,0 +1,63 @@
...
+  min_longitudinal_offset_m_(-rear_overhang_m),
+  max_longitudinal_offset_m_(front_overhang_m + wheel_base_m),
+  min_lateral_offset_m_(-(wheel_tread_m / 2.0 + right_overhang_m)),
+  max_lateral_offset_m_(wheel_tread_m / 2.0 + left_overhang_m),
+  min_height_offset_m_(0.0),
+  max_height_offset_m_(vehicle_height_m)
+{
+}
+
+VehicleInfo from_parameters(rclcpp::Node & node)
```
nnmm: This could potentially also be named `from_parameters` or something similar.

mitsudome-r (approved):

PR #64, opened 2020-10-27
Surround obstacle checker

nnmm:
Interesting parts:
* Because PCL is used, this is a `ament_cmake` package, not `ament_cmake_auto`
* The changes you see in node.cpp with `waitForParam` etc. are using the [`vehicle_info_util`](https://github.com/tier4/Pilot.Auto/pull/58) package
* I changed two `lookupTransform()` calls with timeout to `waitForTransform()`
* I'd like to ask the reviewer to check the vehicle dimensions in `createSelfPolygon()`. I think they were not correct before, but I might be wrong.
* In `getNearestObstacleByPointCloud()`, I removed the dependency on the not-yet-available `pcl_ros` by converting to a `pcl` point cloud first and then transforming.


planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/src/node.cpp
```diff
@@ -180,41 +169,56 @@ void SurroundObstacleCheckerNode::insertStopVelocity(
...
-  geometry_msgs::Pose & pose)
+  const std::string & source, const std::string & target, const rclcpp::Time & time,
+  geometry_msgs::msg::Pose & pose)
 {
   try {
     // get transform from source to target
-    geometry_msgs::TransformStamped ros_src2tgt;
-    ros_src2tgt = tf_buffer_.lookupTransform(source, target, time, ros::Duration(0.1));
-    // convert geometry_msgs::Transform to geometry_msgs::Pose
+    auto tf_future = tf_buffer_.waitForTransform(
```
nnmm, line 177: This is an example where `lookupTransform` was replaced by `waitForTransform`.

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/src/node.cpp
```diff
@@ -391,10 +397,10 @@ bool SurroundObstacleCheckerNode::checkStop(
...
 Polygon2d SurroundObstacleCheckerNode::createSelfPolygon()
 {
-  double front = front_overhang_ + wheel_base_;
-  double rear = -rear_overhang_;
-  double left = wheel_tread_ / 2.0 + left_overhang_ / 2.0;
-  double right = -(wheel_tread_ / 2.0 + right_overhang_ / 2.0);
+  const double front = vehicle_info_.max_longitudinal_offset_m_;
+  const double rear = vehicle_info_.min_longitudinal_offset_m_;
+  const double left = vehicle_info_.max_lateral_offset_m_;
+  const double right = vehicle_info_.min_lateral_offset_m_;
```
nnmm, line 403: `front` and `back` are identical to the old code, but I don't know why `left` and `right` had the `/ 2.0` at the end. The new version seems correct to me:
* `min_lateral_offset_m_` = `-(wheel_tread_m / 2.0 + right_overhang_m)`
* `max_lateral_offset_m_` = `wheel_tread_m / 2.0 + left_overhang_m`.

The names appear wrong to me too (`left` should be the negative number), but I left it like it was.

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/src/node.cpp
```diff
@@ -255,36 +258,39 @@ void SurroundObstacleCheckerNode::getNearestObstacle(
...
-  pcl::PointCloud<pcl::PointXYZ>::Ptr pcl(new pcl::PointCloud<pcl::PointXYZ>);
-  pcl::fromROSMsg(transformed_pointcloud, *pcl);
-
-  for (const auto p : *pcl) {
-    //create boost point
+  Eigen::Affine3f isometry = tf2::transformToEigen(transform_stamped.transform).cast<float>();
+  pcl::PointCloud<pcl::PointXYZ> pcl;
+  pcl::fromROSMsg(*pointcloud_ptr_, pcl);
+  pcl::transformPointCloud(pcl, pcl, isometry);
+  for (const auto p : pcl) {
```
nnmm, line 286: Here are some interesting changes I mentioned:
* While `pcl_ros` transformed between two `sensor_msgs/PointCloud2`, `pcl` can only transform between two `pcl::PointCloud`s, so I convert first. There is only one conversion in the old and new code, so it should have the same performance.
* `pcl_ros` used a matrix, while `pcl` can use an `Affine3f`.
* The pcl docs state that it's allowed to pass in the same point cloud for input and output.

nnmm:
@TakaHoribe The build is only not passing because of the dependency on the not yet merged https://github.com/tier4/Pilot.Auto/pull/58. I have compiled and launched the node locally.

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/src/node.cpp
kenji-miyake: > but I don't know why left and right had the / 2.0 at the end.

This is a bug and fixed in our latest code.

image

mitsudome-r:
vehicle_info_util is already merged.

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/CMakeLists.txt
```diff
@@ -1,68 +1,83 @@
...
+find_package(rclcpp REQUIRED)
+find_package(sensor_msgs REQUIRED)
+find_package(std_msgs REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(tf2_eigen REQUIRED)
+find_package(tf2_ros REQUIRED)
+find_package(visualization_msgs REQUIRED)
+find_package(pcl_conversions REQUIRED)
+find_package(vehicle_info_util REQUIRED)
```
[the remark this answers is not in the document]
nnmm: Done!

nnmm:
@mitsudome-r I rebased this PR on ros2.

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/src/node.cpp
mitsudome-r: I believe we decided not to use waitForTransform if we are asking for the newest transform and just use lookupTransform without timeout, correct?

mitsudome-r:
@nnmm could you check https://github.com/tier4/Pilot.Auto/pull/64#discussion_r518053753?

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/src/node.cpp
nnmm: Yes, good point. I removed it.

PR #65, opened 2020-10-28
Port spline interpolation

cvasfi:
With this PR, spline interpolation package can be built with ROS2 foxy.

### DIFF
--- a/common/math/spline_interpolation/CMakeLists.txt
+++ b/common/math/spline_interpolation/CMakeLists.txt
@@ -1,43 +1,14 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(spline_interpolation)
 
-add_compile_options(-std=c++14)
+if(NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
 
-find_package(catkin REQUIRED COMPONENTS
-  roscpp
-)
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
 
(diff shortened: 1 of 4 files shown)


cvasfi:
@mitsudome-r Shall i pick a reviewer randomly or is there a process defined for that?

common/math/spline_interpolation/CMakeLists.txt
```diff
@@ -1,43 +1,14 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(spline_interpolation)
 
-add_compile_options(-std=c++14)
+if(NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
```
nnmm, line 6: There's some more boilerplate we usually add:
```
if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
  add_compile_options(-Wno-unused-parameter -Wall -Wextra -Wpedantic)
endif()
```
I think Fred has even more in his newest porting doc PR

fred-apex-ai: @nnmm yes, that's true but we can make the standard enforcement stricter in a 2nd step.

@cvasfi please just add the warning flags, and while building temporarily add `-Werror` to make sure there are no  warnings left but don't commit `-Werror`

mitsudome-r: Looks like `Werror` doesn't show any error so I will merge this.

mitsudome-r (approved):

PR #66, opened 2020-10-28
Fix osqp_interface dependencies

### LINKED ISSUE #63: Make osqp_interface use Eigen3 robustly
Follow guidelines of https://github.com/ros2/eigen3_cmake_module#edit-your-cmakeliststxt to avoid subtle errors in include problems for consumers

Assign to @fred-apex-ai

fred-apex-ai:
This PR introduces two fixes:

1. Eigen is consumed and exported properly to fix #63 
2. osqp lib is exported to correct linker errors with static libs

@nnmm This came out our discussion with @jilaada on PR https://github.com/tier4/Pilot.Auto/pull/24

the 2nd point bugged me for a week and I finally better understood what goes on behind the scenes. I will add some guidelines to the porting doc. You can't see it in this PR but in https://github.com/tier4/Pilot.Auto/pull/52 I can now drop this

```diff
modified   control/mpc_follower/CMakeLists.txt
@@ -43,9 +43,6 @@ set(MPC_FOLLOWER_HDR
   )
 
 ament_auto_add_executable(mpc_follower ${MPC_FOLLOWER_SRC} ${MPC_FOLLOWER_HDR})
-# TODO(Frederik.Beaujean) not clear why this isn't taken care of by `ament_auto_add_executable` but `ospq_interface`
-# isn't linked in otherwise
-ament_target_dependencies(mpc_follower osqp_interface)

```
which was needed even though `ament_target_dependencies(mpc_follower osqp_interface)` was already called once because of the linking order of static libs. Pretty nasty!

### DIFF
--- a/common/math/osqp_interface/CMakeLists.txt
+++ b/common/math/osqp_interface/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 find_package(ament_cmake_auto REQUIRED)
 ament_auto_find_build_dependencies()
 
+find_package(eigen3_cmake_module REQUIRED)
 find_package(Eigen3 REQUIRED)
 
 find_package(osqp_vendor REQUIRED)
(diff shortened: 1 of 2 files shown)


common/math/osqp_interface/CMakeLists.txt
```diff
@@ -12,28 +12,28 @@ endif()
...
 
-target_include_directories(osqp_interface PUBLIC "${EXTERNAL_INCLUDE_DIRS}")
-# needed so clients of this package don't need to worry about includes of this package
-ament_export_include_directories("${EXTERNAL_INCLUDE_DIRS}")
+target_include_directories(osqp_interface PUBLIC "${OSQP_INCLUDE_DIR}")
+ament_target_dependencies(osqp_interface Eigen3)
+ament_target_dependencies(osqp_interface osqp_vendor)
+
+# crucial so downstream package builds because osqp_interface exposes osqp.h
+ament_export_include_directories("${OSQP_INCLUDE_DIR}")
```
nnmm, line 35: Doesn't ament_export_dependencies work too?

fred-apex-ai: No, I tried and checked and understand why it fails. First of all, `ament_export_dependencies(osqp_interface)` is called by `ament_auto_package()` below. But it's not enough because it only works with ament-style packages and `osqp` isn't such a package and `osqp_vendor` doesn't expose the necessary info from `osqp`.

An alternative that I just learned today would be to use `ament_export_targets(osqp_interface)`, so I would only have to add the include dir once but hat would mean getting rid of `ament_auto_package()` as far as I understand now. Will check if I get an answer to that question in https://github.com/ament/ament_cmake/issues/292#issuecomment-717967039

fred-apex-ai: Played around a bit but more for my own curiousity. Moving to exporting targets would be another step, and from my point of view this PR is ready to be merged @mitsudome-r

fred-apex-ai:
@nnmm Can you have another look and approve, please?

nnmm (approved):

fred-apex-ai:
@mitsudome-r please merge

PR #69, opened 2020-10-29
ROS2 Porting: dummy_diag_publisher

jilaada:
# dummy_diag_publisher

## Summary

Convert the `dummy_diag_publisher` package to ROS2. This was relatively straightforward apart from the `dynamic_reconfigure` dependency.

Notable changes:
* Additional structs used to represent the original configuration generated using `dynamic_reconfigure`
* Remove fmt and boost deps since they are no longer used
* Add `rqt_reconfigure` to the `package.xml` as an `exec_depend`

## Testing

1. Enter ADE environment and build the workspace up to the package

```
colcon build --packages-up-to dummy_diag_publisher
. install/setup.bash
```

2. Launch files can be launch successfully (make sure that you run rosdep to get the `rqt_reconfigure` package:

```
ros2 launch dummy_diag_publisher dummy_diag_publisher.launch.xml launch_rqt:=true
```

## Discussion Points
* `rqt_runtime_monitor` has not been ported to ROS2, So visualization of these changes are difficult. Are there ROS2 alternatives? I couldn't find anything in my searches.


system/util/dummy_diag_publisher/src/dummy_diag_publisher_node/dummy_diag_publisher_node.cpp
```diff
@@ -14,47 +14,63 @@
...
-  config_ = config;
+  
+  rcl_interfaces::msg::SetParametersResult result;
+  result.successful = true;
+  result.reason = "success";
+
+  DummyDiagPublisherConfig config = config_;
+  try {
+    int status;
+    update_param(parameters, "status", status);
```
mitsudome-r, line 47: `status` must be initialized with original status because update_param only updates when the parameter is actually set.
e.g. if we only set `is_active` parameter in commandline, `status` will have undefined value after update_param.

 ```suggestion
    int status = static_cast<int>(config.status);
    update_param(parameters, "status", status);
```

File: system/util/dummy_diag_publisher/src/dummy_diag_publisher_node/dummy_diag_publisher_node.cpp [hunk 2]
```diff
@@ -63,26 +79,46 @@ void DummyDiagPublisherNode::produceDiagnostics(diagnostic_updater::DiagnosticSt
...
+  configuration_parameters.insert(std::pair<std::string, std::string>("diag_config.msg_error", ""));
+  configuration_parameters.insert(std::pair<std::string, std::string>("diag_config.msg_stale", ""));
 
-  // Dynamic Reconfigure
-  dynamic_reconfigure_.setCallback(boost::bind(&DummyDiagPublisherNode::onConfig, this, _1, _2));
+  this->declare_parameters<std::string>("", configuration_parameters);
+  this->get_parameters("diag_config", configuration_parameters);
+  diag_config_ = DiagConfig(configuration_parameters);
+  
+  // set parameter callback
```
mitsudome-r, line 107: We have need to declare parameter in order to set it from parameter callback.

```suggestion
  // set parameter callback
  config_.status = static_cast<Status>(this->declare_parameter("status", 0));
  config_.is_active = this->declare_parameter("is_active", false);
```

system/util/dummy_diag_publisher/CMakeLists.txt
```diff
@@ -1,69 +1,33 @@
...
-)
-
-# Target
-## dummy_diag_publisher_node
-add_executable(dummy_diag_publisher
+ament_auto_add_executable(${PROJECT_NAME}
   src/dummy_diag_publisher_node/main.cpp
-  src/dummy_diag_publisher_node/dummy_diag_publisher_node.cpp
+  ${DUMMY_DIAG_PUBLISHER_SRC}
+  ${DUMMY_DIAG_PUBLISHER_HEADERS}
```
mitsudome-r, line 26: I think we don't have to add headers when we use ament_auto_add_executables

fred-apex-ai: yes, it's probably my "fault". Adding headers was a recommendation to make life easier for IDEs but I see that at least clion is clever enough and doesn't need it. I will update my porting-doc recommendations. Then you can just write here

```cmake
ament_auto_add_executable(${PROJECT_NAME} src/dummy_diag_publisher_node/main.cpp src/dummy_diag_publisher_node/dummy_diag_publisher_node.cpp)
```

jilaada: Thanks, I'll update this. I noticed that in certain Autoware.Auto packages, the headers were also specified. It would be nice to maybe have these smaller style details ironed out so that all the packages have the same consistent style.

jilaada:
@fred-apex-ai @mitsudome-r I have address the PR comments. Let me know if there are still any issues.

system/util/dummy_diag_publisher/CMakeLists.txt
```diff
@@ -1,69 +1,32 @@
...
+  src/dummy_diag_publisher_node/dummy_diag_publisher_node.cpp)
 
-catkin_package(
-  INCLUDE_DIRS
-    include
-  CATKIN_DEPENDS
-    diagnostic_updater
-    dynamic_reconfigure
-)
+set(DUMMY_DIAG_PUBLISHER_HEADERS
```
mitsudome-r, line 20: Plese remove this since this is not used anymore

jilaada: Done in d869b6f

PR #75, opened 2020-10-30
Fix RCLCPP_INFO_THROTTLE duration unit

TakaHoribe:
The duration arg for  RCLCPP_INFO_THROTTLE() is changed to [ms] from [s] in ROS2.
Some packages are using this function with a very short duration, they seem to be a mistake on porting.

Fix for these packages to use explicit ms duration.


jilaada (approved):
Compiles and code looks as per discussion. LGTM :+1:

PR #81, opened 2020-11-02
[WIP] ROS2 Porting: autoware_state_monitor

jilaada:
# autoware_state_monitor

## Summary

Fairly straightforward package apart from the following remaining ROS1 issues:
 * `topic_tools::ShapeShifter` not currently ported into ROS2 so far

Notable changes:
 * Removal of XmlRpc library - I tried to maintain the templating in `config.h` as much as possible though this might not be the most optimal way of parameter retrieval in ROS2. This includes changes to the configuration yaml layout.
 * Add time to the `StateInput` struct as ROS time is no longer available in non-node classes
 * `rosconsole.h` doesn't seem to be used. I have commented out references to it - if given the ok I will remove it.

## Testing

1. Enter, build and source the workspace
```
ade enter
colcon build --packages-up-to autoware_state_monitor
source install/setup.bash
```

2. Run the example packages
```
ros2 launch autoware_state_monitor autoware_state_monitor.launch.xml
```

Note: Currently the example fails to run properly as one of the member objects is not initialized to completion due to the dependency with `topic_tools::ShapeShifter`.

## Discussion Points
* Removal of `rosconsole.h` - This doesn't seem to be used anywhere, can this be removed?
* New config yaml file structure - I came to this structure to try and maintain as much as possible the templated constructors 

## To Do
 - [x] add work-around for `topic_tools::ShapeShifter`


jilaada:
Update on this package: with the new vendor packages rclcpp_generic and topic_tools we can now use relay node however this package requires ShapeShifter that doesn't require type specification when subscribing to topics. Here is a workaround I thought about:
Add a `type` field which will be used to propagate the subscription creation methods during runtime. This would add a level of overhead maintaining the configuration file - though I don't see the types changing all that much.

jilaada:
@nnmm I may have missed it in previous meetings but was there any further investigation on either porting ShapeShifter or a workaround using the rclcpp_generic implementation? The above comment is a workaround I thought might be good for the time being while alternatives are being investigated - what are your thoughts?

nnmm:
@jilaada Sorry, I meant to reply earlier. While I don't know much about ShapeShifter (do you know what it, it seems like you could use the generic subscription here. You would probably need to find out the type of each message and add it to the yaml like you said.

Taking a step back, is all of this even used? All this callback apparently does is populate `topic_received_time_buffer_` which is private and never read anywhere. There is `calcTopicRate` which seems connected but it is similarly unused. Functionality like that could potentially also use topic statistics: https://index.ros.org/doc/ros2/Concepts/About-Topic-Statistics/. All in all, I would delete it, but you might want to check with the original author.

jilaada:
@nnmm I'll take a look at the built-in topic statistics and maybe just implement the types in the yaml file as I suggested above. Thanks for the suggestion.

jilaada:
@mitsudome-r Just to confirm these are the topics which are monitored and their types:

* /map/vector_map
  * autoware_lanelet2_msgs::msg::MapBin
* /map/pointcloud_map
  * sensor_msgs::msg::PointCloud2
* /sensing/lidar/pointcloud
  * sensor_msgs::msg::PointCloud2
* /sensing/lidar/no_ground/pointcloud
  * sensor_msgs::msg::PointCloud2
* /localization/pose_twist_fusion_filter/pose
  * geometry_msgs::msg::PoseStamped
* /perception/object_recognition/detection/objects
  * autoware_perception_msgs::msg::DynamicObjectWithFeatureArray
* /perception/object_recognition/tracking/objects
  * autoware_perception_msgs::msg::DynamicObjectArray
* /perception/object_recognition/objects
  * autoware_perception_msgs::msg::DynamicObjectArray
  * autoware_perception_msgs::msg::DynamicObjectWithFeatureArray
* /planning/mission_planning/route
  * autoware_planning_msgs::msg::Route
* /planning/scenario_planning/trajectory
  * autoware_planning_msgs::msg::Trajectory
* /control/vehicle_cmd
  * autoware_vehicle_msgs::msg::VehicleCommand
* /vehicle/status/twist
  * geometry_msgs::msg::TwistStamped
* /vehicle/status/steering
  * autoware_vehicle_msgs::msg::Steering
* /system/emergency/control_cmd
  * autoware_control_msgs::msg::ControlCommandStamped

Note: it appears that `/perception/object_recognition/objects` has multiple types depending on a set flag in the launch file: https://github.com/tier4/autoware_launcher.iv.universe/blob/ros2/perception_launch/launch/object_recognition/prediction/prediction.launch.xml#L17-L18. Although I am confident that these types are correct, I found it difficult to follow the perception launch file so please pay close attention to those when reviewing.

jilaada:
@mitsudome-r This is now ready for review. Though some comments: For the above `/perception/object_recognition/objects` topic publishes two possible types of messages. In this implementation I decided to make it so that only the default type is used (`autoware_perception_msgs::msg::DynamicObjectArray`). I investigated a way to use vectors to represent multiple possible types though the issue comes from how to tell that a topic is is a certain type or if it already published. Maybe using something like this: https://github.com/ros2/rcl/blob/foxy/rcl/test/rcl/wait_for_entity_helpers.cpp#L187-L238 as a substitute for the ROS1 `wait_for_message`. Maybe there is a simpler way?

mitsudome-r:
The port looks good to me. 

> For the above /perception/object_recognition/objects topic publishes two possible types of messages. In this implementation I decided to make it so that only the default type is used (autoware_perception_msgs::msg::DynamicObjectArray). 

I think this is okay for now.
We can come back if we face any situation where we really have to support different types.

PR #84, opened 2020-11-04
Port turn signal decider

esteve:


planning/scenario_planning/lane_driving/behavior_planning/turn_signal_decider/src/data_manager.cpp
```diff
@@ -48,56 +48,68 @@ lanelet::ConstLanelets pathToLanes(
...
-    if (!is_pose_ready_) {
-      ROS_WARN_STREAM_THROTTLE(5, ex.what());
-    } else  // if tf suddenly stops comming, then there must be something wrong.
-    {
-      ROS_ERROR_STREAM_THROTTLE(5, ex.what());
+  if (auto node_ptr = node_.lock()) {
+    try {
+      const auto current_time = node_ptr->now();
+      const auto transform =
+        tf_buffer_->lookupTransform("map", "base_link", current_time, rclcpp::Duration(1.0));
```
mitsudome-r, line 89: It seems like it was originally 0.1 seconds for duration.
Is this change intentional?
Also, do you know which constructor is called from double? It looks like most of the constructor is expecting nanoseconds. 
http://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Duration.html#af2ab74e1b3f70436bb0451dd8a17c8e8

It might be more explicit if we use rclcpp::Duration::from_seconds() function

planning/scenario_planning/lane_driving/behavior_planning/turn_signal_decider/src/data_manager.cpp
```diff
@@ -107,10 +119,14 @@ bool DataManager::isPoseValid() const
...
   }
 
-  // check time stamp
-  constexpr double timeout = 1.0;
-  if (ros::Time::now() - vehicle_pose_.header.stamp > ros::Duration(timeout)) {
-    return false;
+  if (auto node_ptr = node_.lock()) {
+    // check time stamp
+    constexpr double timeout = 1.0;
+    if (node_ptr->now() - vehicle_pose_.header.stamp > rclcpp::Duration(timeout)) {
```
mitsudome-r, line 125: same here with the units.

mitsudome-r (commented):
Looks good overall, but with few comments about Duration.
I may be wrong, but please take a look

planning/scenario_planning/lane_driving/behavior_planning/turn_signal_decider/src/data_manager.cpp
esteve: Fixed in 7270016

planning/scenario_planning/lane_driving/behavior_planning/turn_signal_decider/src/data_manager.cpp
esteve: Good catch, I copied that Duration from elsewhere and forgot to update the value, thanks.

Fixed in 7270016

PR #85, opened 2020-11-04
Update osqp_vendor dependency

esteve:


esteve:
I'm merging this because the change is very minor and CI passes.

PR #86, opened 2020-11-04
Port obstacle stop planner

sgermanserrano:
Port of `obstacle-stop-planner` to ROS2. A couple of notes:

- In the `CMakeLists.txt` compile option `-O3` needs to be added for the `static constexpr` in the `adaptive_cruise_control.hpp` parameters to be compiled properly.
- In `node.cpp` line [212](https://github.com/tier4/Pilot.Auto/pull/86/files#diff-9493ac6ff50c82cecb25fb47d2f6fb1a6a512d07e8aad7d2693d94cc419a1107R212) I've replaced the `pcl_ros::transformPointCloud` usage with the `pcl::transformPointCloud` equivalent as `pcl_ros` is not available in ROS2 yet

### DIFF
--- a/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/adaptive_cruise_control.cpp
+++ b/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/adaptive_cruise_control.cpp
@@ -31,19 +31,19 @@ using Line = bg::model::linestring<Point>;
 
 namespace
 {
-Point convertPointRosToBoost(const geometry_msgs::Point & point)
+Point convertPointRosToBoost(const geometry_msgs::msg::Point & point)
 {
   const Point point2d(point.x, point.y);
   return point2d;
 }
 
-geometry_msgs::Vector3 rpyFromQuat(const geometry_msgs::Quaternion & q)
+geometry_msgs::msg::Vector3 rpyFromQuat(const geometry_msgs::msg::Quaternion & q)
 {
   tf2::Quaternion quat(q.x, q.y, q.z, q.w);
   tf2::Matrix3x3 mat(quat);
   double roll, pitch, yaw;
   mat.getRPY(roll, pitch, yaw);
-  geometry_msgs::Vector3 rpy;
+  geometry_msgs::msg::Vector3 rpy;
   rpy.x = roll;
   rpy.y = pitch;
   rpy.z = yaw;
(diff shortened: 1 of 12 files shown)


planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/adaptive_cruise_control.cpp
```diff
@@ -106,38 +106,14 @@ constexpr double sign(const double value)
...
-    const auto result = nh.getParam(key, value);
-    if (result) {
-      return value;
-    }
-    ROS_WARN("waiting for parameter `%s` ...", key.c_str());
-    rate.sleep();
-  }
-}
-
-}  // namespace
```
simon-t4, line 132: Are these two function not used at all?

planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/node.cpp
```diff
@@ -27,45 +26,19 @@
...
 #include <vector>
 #define EIGEN_MPL2_ONLY
-#include <Eigen/Core>
-#include <Eigen/Geometry>
+#include <eigen3/Eigen/Core>
+#include <eigen3/Eigen/Geometry>
 namespace
 {
-template <class T>
-T getParam(const ros::NodeHandle & nh, const std::string & key, const T & default_value)
```
simon-t4, line 35: Ahh, I see these functions seem to left over from a common code source?

planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/node.cpp
```diff
@@ -225,12 +209,13 @@ void ObstacleStopPlannerNode::pathCallback(
 
     Eigen::Matrix4f affine_matrix =
       tf2::transformToEigen(transform_stamped.transform).matrix().cast<float>();
-    sensor_msgs::PointCloud2 transformed_obstacle_ros_pointcloud;
-    pcl_ros::transformPointCloud(
-      affine_matrix, *obstacle_ros_pointcloud_ptr_, transformed_obstacle_ros_pointcloud);
+    pcl::PointCloud<pcl::PointXYZ>::Ptr obstacle_ros_pointcloud_pcl_ptr_(
```
simon-t4, line 212: Should this variable name exclude "ros" as it is no longer a ros type?

simon-t4:
@sgermanserrano Hi, a couple of comments above, but in general the port seems straight forward and well done.

planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/adaptive_cruise_control.cpp
sgermanserrano: yeah, they are removed as part of updating how the parameters are handled in ROS2

planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/node.cpp
sgermanserrano: they were used to handle parameters so we don't need them anymore

planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/node.cpp
sgermanserrano: you are right, I've updated the variable name in the latest commit

PR #87, opened 2020-11-04
Update logger name for SOR class

sgermanserrano:
Updated logger name when called within the SOR class


mitsudome-r (approved):
Thanks for the fix!

PR #90, opened 2020-11-04
Port dummy_perception_publisher to ROS2

nnmm:
I learned a few small things that I'll put into the porting doc, but porting this package was relatively straightforward. I replaced `uuid_msgs` and with `unique_identifier_msgs` and removed the unused `unique_id` dependency.


cvasfi:
@nnmm I have skimmed the code changes and don't see anything wrong My only comment would be that using `ament_auto_cmake` would simplify the CMake file. Though that's a *take it or leave it* kind of comment as I see that many other packages also not use it.

nnmm:
@cvasfi I had problems with `ament_cmake_auto` [for message definitions](https://github.com/ament/ament_cmake/issues/301), so I didn't use it here.

If you don't think any changes are needed, you can approve the PR (I'll still need to find out why some lint test failed). In case you're new to GitHub, you can do this by going to the "Files changed" view and select "Approve" in the green "Review changes" dropdown.

cvasfi (approved):

cvasfi:
@nnmm 
Fair enough. I approved.

mitsudome-r:
@nnmm CI for tests is failing.
It looks like you added lint test, but I think it was failing from ROS1 code.
I will leave up to you if you want to do the fix now, or if you want to remove test out and do it after initial porting.

nnmm:
@mitsudome-r Yes, I had never enabled lint tests before, and I learned that for instance `cppcheck` requires the copyright notice to use `//` style comments and not `/*`. They are all fixed now.

PR #95, opened 2020-11-09
Port map based prediction

sgermanserrano:
Port of `map_based_prediction` to ROS2.

I've added a `toHexString` function [here](https://github.com/tier4/Pilot.Auto/pull/95/files#diff-6609d065fa15d5b881b14c3a7b45a522484085c047f05d28f884ca8e27dda4b3R44) as it seems `unique_id::toHexString` is not available.

### DIFF
--- a/perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
+++ b/perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
@@ -4,9 +4,9 @@
   <arg name="prediction_time_horizon" default="10" />
   <arg name="prediction_sampling_delta_time" default="0.5" />
   <arg name="vector_map_topic" default="/map/vector_map" />
-  <node pkg="map_based_prediction" type="map_based_prediction" name="map_based_prediction" output="screen">
-    <param name="prediction_time_horizon" value="$(arg prediction_time_horizon)" />
-    <param name="prediction_sampling_delta_time" value="$(arg prediction_sampling_delta_time)" />
-    <remap from="/vector_map" to="$(arg vector_map_topic)"/>
+  <node pkg="map_based_prediction" exec="map_based_prediction" name="map_based_prediction" output="screen">
+    <param from="$(var prediction_time_horizon)" />
+    <param from="$(var prediction_sampling_delta_time)" />
+    <remap from="/vector_map" to="$(var vector_map_topic)"/>
   </node>
 </launch>
--- a/perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
+++ b/perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" ?>
 
 <launch>
-  <arg name="prediction_time_horizon" default="10" />
+  <arg name="prediction_time_horizon" default="10.0" />
   <arg name="prediction_sampling_delta_time" default="0.5" />
   <arg name="vector_map_topic" default="/map/vector_map" />
   <node pkg="map_based_prediction" exec="map_based_prediction" name="map_based_prediction" output="screen">
-    <param from="$(var prediction_time_horizon)" />
-    <param from="$(var prediction_sampling_delta_time)" />
+    <param name="$(var prediction_time_horizon)" />
+    <param name="$(var prediction_sampling_delta_time)" />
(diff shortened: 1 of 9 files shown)


esteve (approved):

perception/object_recognition/prediction/map_based_prediction/src/map_based_prediction_ros.cpp
```diff
@@ -47,8 +41,18 @@
 #include "map_based_prediction.h"
 #include "map_based_prediction_ros.h"
 
+std::string toHexString(const unique_identifier_msgs::msg::UUID & id){
+  auto begin = id.uuid.begin();
+  auto end = id.uuid.end();
+  std::ostringstream hex;
+  hex << std::hex;
```
esteve, line 48: This will not pad integers with a 0 to ensure that their length is 2. See https://github.com/tier4/Pilot.Auto/blob/ros2/perception/util/visualizer/dynamic_object_visualization/include/dynamic_object_visualization/dynamic_object_visualizer.h#L66-L70 for example how I re-implemented `toHexString`, perhaps this should go into the `unique_identifier_msgs` package

esteve (changes requested):

perception/object_recognition/prediction/map_based_prediction/src/map_based_prediction_ros.cpp
sgermanserrano: @esteve if there are already 2 nodes sharing the functionality it definitively seems like it should be common, I've seen also in [autoware_perception_rviz_plugins](https://github.com/tier4/Pilot.Auto/blob/master/common/util/rviz_plugins/autoware_perception_rviz_plugin/src/tools/car_pose.cpp#L137) that there are more UUID-related functions that seem to be commonly used

sgermanserrano: @esteve  I've updated the function in the latest commit

perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
```diff
@@ -4,9 +4,9 @@
   <arg name="prediction_time_horizon" default="10" />
   <arg name="prediction_sampling_delta_time" default="0.5" />
   <arg name="vector_map_topic" default="/map/vector_map" />
-  <node pkg="map_based_prediction" type="map_based_prediction" name="map_based_prediction" output="screen">
-    <param name="prediction_time_horizon" value="$(arg prediction_time_horizon)" />
-    <param name="prediction_sampling_delta_time" value="$(arg prediction_sampling_delta_time)" />
-    <remap from="/vector_map" to="$(arg vector_map_topic)"/>
+  <node pkg="map_based_prediction" exec="map_based_prediction" name="map_based_prediction" output="screen">
+    <param from="$(var prediction_time_horizon)" />
+    <param from="$(var prediction_sampling_delta_time)" />
```
mitsudome-r, line 9: These are not yaml files so `from` is not appropriate.

```suggestion
    <param name="prediction_time_horizon" value="$(var prediction_time_horizon)" />
    <param name="prediction_sampling_delta_time" value="$(var prediction_sampling_delta_time)" />
```

perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
```diff
@@ -4,9 +4,9 @@
   <arg name="prediction_time_horizon" default="10" />
```
mitsudome-r, line 4: Please change this to `10.0` since ROS2 is more strict about handing double value if it is declared as double parameter.

```suggestion
  <arg name="prediction_time_horizon" default="10.0" />
```

mitsudome-r (commented):
Looks generally good to me, but with few comments about the launch file.

perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
sgermanserrano: added in the latest commit

perception/object_recognition/prediction/map_based_prediction/launch/map_based_prediction.launch.xml
sgermanserrano: fixed in the latest commit

mitsudome-r (approved):
LGTM

PR #96, opened 2020-11-09
Revert "Add geometry2 to repos (#76)"

nnmm:
https://github.com/ros/rosdistro/pull/27177 is now available through the ROS package repositories: I upgraded my system in ADE today and got 

```
$ apt show ros-foxy-tf2
Package: ros-foxy-tf2
Version: 0.13.6-1focal.20201028.172335
```

which is the version that has the PR required for `ndt_scan_matcher`.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -28,6 +28,10 @@ jobs:
         rosdep update
         DEBIAN_FRONTEND=noninteractive rosdep install --from-paths . --ignore-src --rosdistro foxy -y
 
+    - name: Show ros-foxy-tf2 version
+      run: |
+        apt show ros-foxy-tf2
+
     - name: Build
       run: |
         . /opt/ros/foxy/setup.sh
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -28,10 +28,6 @@ jobs:
         rosdep update
         DEBIAN_FRONTEND=noninteractive rosdep install --from-paths . --ignore-src --rosdistro foxy -y
 
-    - name: Show ros-foxy-tf2 version
-      run: |
-        apt show ros-foxy-tf2
-
     - name: Build
       run: |
         . /opt/ros/foxy/setup.sh
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -24,9 +24,14 @@ jobs:
 
     - name: Install missing dependencies
       run: |
-        sudo apt update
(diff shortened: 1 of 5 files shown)


nnmm:
@esteve Do you know why CI doesn't appear to use that package yet?

esteve:
@nnmm hard to say, but I think it could be because the tf packages were removed from the package.xml files.

nnmm:
@esteve I don't think that's it, CI failed even after you re-added them and `ndt_scan_matcher` builds without problems in a fresh ADE environment on my local machine at commit `30ffda7` (this PR before you added tf2 dependencies).

I have pushed a temporary change to CI that prints the version fo the installed package to see if it's the correct one.

nnmm:
Okay, it's `0.13.6-1focal.20201028.17233`, which should be right.

mitsudome-r:
I think it is because we are using ros:foxy tag for docker used in the CI.
The docker image is updated 19 days ago which is before the geometry2 package was released.

I think adding `apt update && apt upgrade` command in build_and_test.yaml would solve the error.

esteve:
@nnmm @mitsudome-r alternatively we could just install the updated version without upgrading the whole system:

`sudo apt-get install - y ros-foxy-tf2-ros>=0.13.6-1focal.20201028.17233`

sgermanserrano:
@esteve @nnmm @mitsudome-r I've been doing some digging with the latest `ros:foxy` docker image locally. It seems that the upgrade is needed for a number of packages, the issue with the failing build is that `tf2_geometry_msgs/tf2_geometry_msgs.h` does not contain the declaration for `tf2::doTransform` with `geometry_msgs::msg::PoseWithCovarianceStamped` messages whereas doing `upgrade` moves everything forward

esteve:
@nnmm I've listed all the packages and their versions in https://github.com/tier4/Pilot.Auto/pull/96/commits/62269910a8dd6064b90724614edef1e8b650c780, I've tested this locally with the `ros:foxy` Docker image and it build fine now.

nnmm:
@esteve Great that this works. Just curious, why don't we `sudo apt upgrade` all packages?

esteve:
@nnmm mainly to avoid CI taking too long and also so that we know exactly what version of the Docker image we're using without deviating too much. In any case, I'd prefer to use ADE in the CI, since we can avoid issues like this by having the same environment in development and in testing.

esteve:
@nnmm I've pushed 2bc259e to clean up the `colcon test` invocation, we're ready to merge this PR after CI passes.

esteve (approved):

PR #99, opened 2020-11-12
Ros2 port traffic light map

simon-t4:
Porting traffic light map detector module to ROS2.

Module has standard ROS interfaces, port:
[x] CMakeLists to use ament
[x] convert messages, rclcpp calls
[x] setup subscribers and publishers

### DIFF
--- a/perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
+++ b/perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
@@ -19,8 +19,6 @@ find_package(Eigen3 REQUIRED)
 
 include_directories(
   ${EIGEN3_INCLUDE_DIR}
-  include
-  src/autoware/pilot.auto/map/lanelet2_extension/include
 )
 
 ament_auto_add_executable(traffic_light_map_based_detector_node
--- a/perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
+++ b/perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
@@ -10,6 +10,7 @@ endif()
 
 
 find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
 
 find_package(Eigen3 REQUIRED)
 
--- a/perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
+++ b/perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
@@ -1,40 +1,24 @@
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.5)
 project(traffic_light_map_based_detector)
 
-add_compile_options(-std=c++14)
+if(NOT CMAKE_CXX_STANDARD)
+	set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic -Wno-unused-parameter)
+endif()
 
-find_package(catkin REQUIRED COMPONENTS
-  autoware_perception_msgs
-  autoware_planning_msgs
-  geometry_msgs
-  lanelet2_extension
-  roscpp
-  sensor_msgs
-  tf2
-  tf2_ros
-)
+
+find_package(ament_cmake_auto REQUIRED)
+
 find_package(Eigen3 REQUIRED)
 
-catkin_package(
-  INCLUDE_DIRS include
-  CATKIN_DEPENDS
-    autoware_perception_msgs
-    autoware_planning_msgs
-    geometry_msgs
-    lanelet2_extension
-    roscpp
-    sensor_msgs
-    tf2
-    tf2_ros
-)
 
 ###########
 ## Build ##
 ###########
 
 include_directories(
-  include
-  ${catkin_INCLUDE_DIRS}
   ${EIGEN3_INCLUDE_DIR}
 )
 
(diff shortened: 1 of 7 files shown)


simon-t4:
@mitsudome-r One issue is that roi vibration parameters need a default value. I have assigned 0.0 default values but I ma not sure this is appropriate.

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
```diff
@@ -74,11 +88,12 @@ void MapBasedDetector::cameraInfoCallback(const sensor_msgs::CameraInfo::ConstPt
     camera_pose_stamped.pose.orientation.z = transform.transform.rotation.z;
     camera_pose_stamped.pose.orientation.w = transform.transform.rotation.w;
   } catch (tf2::TransformException & ex) {
-    ROS_WARN_THROTTLE(5, "cannot get transform frome map frame to camera frame");
+    RCLCPP_WARN_THROTTLE(get_logger(), *get_clock(), 5,
```
mitsudome-r, line 91: It seems the units for THROTTLE has changed to ms from seconds so please change this to `5000`

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
```diff
@@ -375,15 +399,15 @@ void MapBasedDetector::publishVisibleTrafficLights(
     point.z = tf_camera2tl.getOrigin().z();
     marker.points.push_back(point);
 
-    marker.lifetime = ros::Duration(0.2);
+    marker.lifetime = rclcpp::Duration(0.2);
```
mitsudome-r, line 402: I think this is implicitly calling constructor with nanosec.
You can use `rclcppp::Duration::from_seconds(0.2)` function instead.

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
```diff
@@ -17,54 +17,68 @@
...
+  autoware_perception_msgs::msg::TrafficLightRoiArray output_msg;
   output_msg.header = camera_info_ptr_->header;
-  geometry_msgs::PoseStamped camera_pose_stamped;
+  geometry_msgs::msg::PoseStamped camera_pose_stamped;
   try {
-    geometry_msgs::TransformStamped transform;
+    geometry_msgs::msg::TransformStamped transform;
     transform = tf_buffer_.lookupTransform(
-      "map", input_msg->header.frame_id, input_msg->header.stamp, ros::Duration(0.2));
+       "map", input_msg->header.frame_id, tf2::TimePointZero);
```
mitsudome-r, line 81: I think we've found out that lookupTransform with timeout works if timestamp is not zero so could you change this to use the exact time with timeout?
e.g. 
```
tf2_buffer_.lookupTransform(target_frame, source_frame, time_stamp, rclcpp::Duration::from_seconds(0.2));
```

perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
```diff
@@ -1,69 +1,38 @@
...
 ###########
 ## Build ##
 ###########
 
 include_directories(
-  include
-  ${catkin_INCLUDE_DIRS}
   ${EIGEN3_INCLUDE_DIR}
+  include
+  src/autoware/pilot.auto/map/lanelet2_extension/include
```
mitsudome-r, line 23: These two lines are not necessary.

mitsudome-r (commented):
It looks good overall with just a few comments.

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
simon-t4: Do we have a working example? BufferCore doesn't seem to accept such a function. Should we switch to back Buffer?

perception/traffic_light_recognition/traffic_light_map_based_detector/CMakeLists.txt
simon-t4: Thanks, removed

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
simon-t4: Thanks, done.

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
simon-t4: Thanks, done.

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
simon-t4: @mitsudome-r I have pushed the above changes, excluding the lookuptransform. Let me know if you want to to revisit this.

perception/traffic_light_recognition/traffic_light_map_based_detector/launch/traffic_light_map_based_detector.launch.xml
```diff
@@ -7,11 +7,10 @@
...
-    <remap from="~input/vector_map" to="$(arg input/vector_map)" />
-    <remap from="~input/camera_info" to="$(arg input/camera_info)" />
-    <remap from="~input/route" to="$(arg input/route)" />
-    <remap from="~output/rois" to="$(arg output/rois)" />
-    <rosparam command="load" file="$(find traffic_light_map_based_detector)/config/traffic_light_map_based_detector.yaml" />
+  <node pkg="traffic_light_map_based_detector" exec="traffic_light_map_based_detector_node" name="traffic_light_map_based_detector" output="screen">
+    <remap from="/input/vector_map" to="$(val input/vector_map)" />
+    <remap from="/input/camera_info" to="$(val input/camera_info)" />
+    <remap from="/input/route" to="$(val input/route)" />
+    <remap from="/output/rois" to="$(val output/rois)" />
```
mitsudome-r, line 14: `val` must be `var`

mitsudome-r (commented):
I noticed one more small fix in launch file.

simon-t4:
@mitsudome-r Done!

perception/traffic_light_recognition/traffic_light_map_based_detector/src/node.cpp
cvasfi: > I think we've found out that lookupTransform with timeout works if timestamp is not zero so could you change this to use the exact time with timeout?
@mitsudome-r  Strangely that is only the case with `waitForTransform` I think, reading the logic of `tf2_ros::lookupTransform`, it seems like it should work.

I am still a bit confused with what's the issue here so here's my take:

* If the aim is to have `lookupTransform` wait until the transform is available, then using `tf2_ros::Buffer::lookupTransform` with the message timestamp + timeout should be used.
* If the aim is to work with the current state of the buffer and not wait for anything then:
   * Use lookupTransform with exact time stamp -> You might not succeed if the data has not arrived just yet.
   * Or do something more fail-safe:

```cpp
if(tf_buffer_.can_transform(/* */, time_stamp)){
   tf = tf_buffer_. lookupTransform(/* */, time_stamp);
} else {  // Data hasn't arrived yet, use the best we got:
   tf = tf_buffer_. lookupTransform(/* */, tf2::TimePointZero);
}
```

simon-t4: @cvasfi @mitsudome-r 
Thanks for the suggestions.
Used `tf2_ros::Buffer` and `lookupTransform` with message timestamp and timeout.

Something not obvious (to me), I had to instantiate the Buffer object with the node clock or else there is a compilation error in the TransformListener instantiation.

mitsudome-r (approved):

PR #101, opened 2020-11-13
ROS2 Porting: scenario_selector

jilaada:
## Summary

Fairly straightforward port of the scenario_selector package - dependencies to topic tools package so ensure that the `ros2` branch of AutowareArchitectureProposal is up to date. 

Notable Changes:
* Remove boost dependencies
  * `boost::function` -> `std::function`
  * `boost::bind` -> `std::bind`
* Pull in `topic_tools` dependencies

## Testing

Compilation should be successful and launch files should launch with issue. The launch files here do not come with default parameters so you will need to input these manually during launch, i.e:

```
ros2 launch scenario_selector dummy_scenario_selector_parking.launch.xml output_scenario:=test_out_scenario input_parking_trajectory:=test_in_traj output_trajectory:=test_out_traj
```

## Discussion Points

* Launch rostopic as a node in the launch file has been commented out - I am not sure if there is a different way of achieving the publishing of topics via the launch file.
* `fromMapBin` now only takes in two argument - is this correct?

### DIFF
--- a/planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
+++ b/planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
@@ -9,6 +9,10 @@
   <arg name="input_twist" default="" />
   <arg name="is_parking_completed" default="" />
 
-  <node pkg="topic_tools" type="relay" name="scenario_trajectory_relay" args="$(arg input_lane_driving_trajectory) $(arg output_trajectory)" output="screen"/>
-  <node pkg="rostopic" type="rostopic" name="scenario_pub" args="pub $(arg output_scenario) autoware_planning_msgs/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'" output="screen"/>
+  <node pkg="topic_tools" exec="relay" name="scenario_trajectory_relay" output="log" >
+    <param name="input_topic" value="$(var input_lane_driving_trajectory)" />
+    <param name="output_topic" value="$(var output_trajectory)" />
+    <param name="type" value="autoware_planning_msgs/msg/Trajectory" />
+  </node>
+  <!-- <node pkg="ros2topic" type="ros2topic" name="scenario_pub" args="pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'" output="screen"/> -->
 </launch>
--- a/planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
+++ b/planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
@@ -15,6 +15,6 @@
     <param name="type" value="autoware_planning_msgs/msg/Trajectory" />
   </node>
 
-  <executable cmd="ros2 topic pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'"
-              name="scenario_pub" output="screen" shell="true"/>
+  <arg name="cmd" default="ros2 topic pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'" />
+  <executable cmd="$(var cmd)" name="scenario_pub" output="screen" shell="true"/>
 </launch>
--- a/planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
+++ b/planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
@@ -14,5 +14,7 @@
     <param name="output_topic" value="$(var output_trajectory)" />
     <param name="type" value="autoware_planning_msgs/msg/Trajectory" />
   </node>
-  <!-- <node pkg="ros2topic" type="ros2topic" name="scenario_pub" args="pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'" output="screen"/> -->
+
+  <executable cmd="ros2 topic pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'"
+              name="scenario_pub" output="screen" shell="true"/>
 </launch>
--- a/planning/scenario_planning/scenario_selector/src/scenario_selector_node/scenario_selector_node.cpp
+++ b/planning/scenario_planning/scenario_selector/src/scenario_selector_node/scenario_selector_node.cpp
@@ -33,15 +33,15 @@
 namespace
 {
 template <class T>
-void onData(const T & data, T * buffer)
+void onData(const T data, T * buffer)
 {
   *buffer = data;
 }
 
 template <class T>
-boost::function<void(const T &)> createCallback(T * buffer)
+std::function<void(T)> createCallback(T * buffer)
 {
-  return static_cast<boost::function<void(const T &)>>(boost::bind(onData<T>, _1, buffer));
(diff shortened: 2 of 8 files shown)


jilaada:
CI is failing as topic_tools is not part of the build_depends.repos file I think

planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
```diff
@@ -9,6 +9,10 @@
...
   <arg name="is_parking_completed" default="" />
 
-  <node pkg="topic_tools" type="relay" name="scenario_trajectory_relay" args="$(arg input_lane_driving_trajectory) $(arg output_trajectory)" output="screen"/>
-  <node pkg="rostopic" type="rostopic" name="scenario_pub" args="pub $(arg output_scenario) autoware_planning_msgs/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'" output="screen"/>
+  <node pkg="topic_tools" exec="relay" name="scenario_trajectory_relay" output="log" >
+    <param name="input_topic" value="$(var input_lane_driving_trajectory)" />
+    <param name="output_topic" value="$(var output_trajectory)" />
+    <param name="type" value="autoware_planning_msgs/msg/Trajectory" />
+  </node>
+  <!-- <node pkg="ros2topic" type="ros2topic" name="scenario_pub" args="pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: LaneDriving, activating_scenarios: [LaneDriving]}'" output="screen"/> -->
```
cvasfi, line 17: I played out a little bit and i think the following should work given that you add `ros2cli` as a dependency:

```xml
  <executable cmd="ros2 topic pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: Parking, activating_scenarios: [Parking]}'"
              name="scenario_pub" output="screen" shell="true"/>
```

cvasfi (changes requested):

planning/scenario_planning/scenario_selector/src/scenario_selector_node/scenario_selector_node.cpp
```diff
@@ -33,15 +31,15 @@
 namespace
 {
 template <class T>
-void onData(const T & data, T * buffer)
+void onData(const T data, T * buffer)
```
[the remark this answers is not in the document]
jilaada: Ah my rationale was that since other subscription callbacks have removed the explicit `&` when using `ConstSharedPtr` I would keep the same structure here. But you have a good point and since this is not like other subscription callbacks it should be explicit. I have made the fix here: fa49bfd

planning/scenario_planning/scenario_selector/src/scenario_selector_node/scenario_selector_node.cpp
```diff
@@ -281,58 +280,77 @@ void ScenarioSelectorNode::onTimer(const ros::TimerEvent & event)
...
-  private_nh_.param<double>("th_stopped_velocity_mps", th_stopped_velocity_mps_, 0.01);
-
-  // Input
-  input_lane_driving_.sub_trajectory = private_nh_.subscribe(
-    "input/lane_driving/trajectory", 1, createCallback(&input_lane_driving_.buf_trajectory));
+  update_rate_ = this->declare_parameter<double>("update_rate", 10.0);
+  th_max_message_delay_sec_ = this->declare_parameter<double>("th_max_message_delay_sec", 1.0);
+  th_arrived_distance_m_ = this->declare_parameter<double>("th_arrived_distance_m", 1.0);
+  th_stopped_time_sec_ = this->declare_parameter<double>("th_stopped_time_sec", 1.0);
+  th_stopped_velocity_mps_ = this->declare_parameter<double>("th_stopped_velocity_mps", 0.01);
```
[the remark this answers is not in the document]
jilaada: Thanks for pointing this out: fix here fa49bfd

planning/scenario_planning/scenario_selector/src/scenario_selector_node/scenario_selector_node.cpp
```diff
@@ -211,68 +210,68 @@ autoware_planning_msgs::Scenario ScenarioSelectorNode::selectScenario()
...
   return scenario;
 }
 
-void ScenarioSelectorNode::onMap(const autoware_lanelet2_msgs::MapBin & msg)
+void ScenarioSelectorNode::onMap(const autoware_lanelet2_msgs::msg::MapBin::ConstSharedPtr msg)
 {
   lanelet_map_ptr_ = std::make_shared<lanelet::LaneletMap>();
-  lanelet::utils::conversion::fromBinMsg(
-    msg, lanelet_map_ptr_, &traffic_rules_ptr_, &routing_graph_ptr_);
+  lanelet::utils::conversion::fromBinMsg(*msg, lanelet_map_ptr_);
```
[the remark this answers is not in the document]
jilaada: Oh sorry about this, for some reason I could not find references to this! That was very strange. Anyway I see could correctly compile the call to the method with 4 arguments and have made the correction here: fa49bfd

I actually made this same mistake in another package so thanks for picking this up. I made a fix for it there as well but for some reason I cannot set you as a reviewer for that repository - you are tagged though so please review it when you have the time.

planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
jilaada: So I tried this and added the line to the dummy launch files here: fa49bfd

However when running these packages with the following command I get an error: 
```
INFO] [scenario_pub-2]: process started with pid [4947]
[scenario_pub-2] usage: ros2 [-h] Call `ros2 <command> -h` for more detailed usage. ...
[scenario_pub-2] ros2: error: unrecognized arguments: Parking, activating_scenarios: [Parking]}
[ERROR] [scenario_pub-2]: process has died [pid 4947, exit code 2, cmd 'ros2 topic pub /test_out_scenario autoware_planning_msgs/msg/Scenario {current_scenario: Parking, activating_scenarios: [Parking]}'].
```

Not too sure what this might indicate but running the exact same command directly through the command line runs the command line publisher correctly, but through ros2 launch this seems to fail. 

So then I did some experimenting and found that by simply moving the value of `cmd` into its own argument as so:
```
<arg name="cmd" default="ros2 topic pub $(var output_scenario) autoware_planning_msgs/msg/Scenario '{current_scenario: Parking, activating_scenarios: [Parking]}'" />
<executable cmd="$(var cmd)" name="scenario_pub" output="screen" shell="true"/>
```

This way I was able to successfully publish to the topic. I don't really have an explanation for why this is the case. Changes here: c0b6d61

jilaada:
@cvasfi Thanks for the thorough review and the excellent suggestions! I fixed the changes you requested and now CI should pass as #102 has been merged.

mitsudome-r (approved):
LGTM

planning/scenario_planning/scenario_selector/launch/dummy_scenario_selector_lane_driving.launch.xml
cvasfi: That's strange, I think i could actually launch it directly but I don't have a problem with this state as well.

planning/scenario_planning/scenario_selector/src/scenario_selector_node/scenario_selector_node.cpp
```diff
@@ -39,9 +37,9 @@ void onData(const T & data, T * buffer)
 }
 
 template <class T>
-boost::function<void(const T &)> createCallback(T * buffer)
+std::function<void(T)> createCallback(T * buffer)
 {
-  return static_cast<boost::function<void(const T &)>>(boost::bind(onData<T>, _1, buffer));
+  return static_cast<std::function<void(T)>>(std::bind(&onData<T>, std::placeholders::_1, buffer));
```
cvasfi, line 42: I think this should be `std::function<void(const T &)>>`, right? Or is it equivalent?

cvasfi: Oh i didn't realize this was merged, just nevermind this comment.

PR #107, opened 2020-11-16
Removed osqp and ublox from build_depends.repos

esteve:
Remove osqp and ublox now that they have been released to the ROS buildfarm.

### DIFF
--- a/build_depends.repos
+++ b/build_depends.repos
@@ -1,12 +1,4 @@
 repositories:
-  vendor/osqp:
-    type: git
-    url: https://github.com/tier4/osqp_vendor.git
-    version: 0.0.2
-  vendor/ublox:
-    type: git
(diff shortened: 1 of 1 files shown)


nnmm (approved):

fred-apex-ai (approved):
@esteve looks good. What do I have to do to clean my environment from the old installation and the get the dependencies from build barn to make sure it works now?

esteve:
@fred-apex-ai did you install osqp from source? rosdep install - - from . - - ignore-src should install any missing dependencies

nnmm:
@fred-apex-ai I think build_depends.repos doesn't affect you, only CI. For cleaning up, I did `rm -rf src/vendor` and then repeated the `vcs import` command from the porting docs, which is probably not the best solution but worked.

PR #114, opened 2020-11-19
Port points preprocessor

TakaHoribe:
## TODO

 - [x] filter base class
 - [x] filters implementation
   - [x] compare_map
   - [x] concatnate_data (not inherits the filter class)
   - [x] crop_box
   - [x] downsample
   - [x] ground
   - [x] outlier
   - [x] passthrough
   - [x] pointcloud_accumulator
 - [x] DynamicReconfigure - use service for all filters
 - [x] create launch files

## stderr when build

THis should be fixed.

```
--- stderr: pointcloud_preprocessor                                               
CMake Warning (dev) at CMakeLists.txt:18 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  CMake variable PCL_ROOT is set to:

    /usr

  For compatibility, CMake is ignoring the variable.
This warning is for project developers.  Use -Wno-dev to suppress it.

---
```

### Filter::subscribe() and Filter::onInit()
The subscribe() and onInit() functions in the Filter and its parent class have been moved into the Filter constructor. If the child class requires other subscribers, it will be implemented in the constructor of the child class.

### Filter::unsubscribe()
Removed. This was for the NodeletLazy to stop subscribe when the published topic is not received by any nodes. 

### Filter::child_init()
The reconfigure was set in this function. Moved to paramCallback() defined in each child filter classes. 

### Filter::Filter()
Add filter_name and node_option in the constructor for the component initialization. 

### ConcatenateData Adaptive Timer Rate
The function to change the timer rate in runtime is removed since the ROS2 generic timer doesn't have the API to change the timer rate after initialization. This changes the logic and performance. Need to create an issue to handle this.
- [x] create an issue for the adaptive timer rate 

### pcl_nodelet
Essential functions are combined to Filter class. Other utils such as lazy-function are removed.

### waitForTransform()
Replaced with tf_future. (The waitForTransform can be used here since the specific timestamp is used to get the tf)

### tf_listener and tf_buffer in child filter classes
Moved to the parent Filter class.

### Dynamic Reconfigure
Replaced to the parameter service callback.

### message filter
The porting was almost straightforward, but the PCL topic is not supported. The message filter for the PCL is removed since it was not used anywhere.

### deprecated pcl functions
 - replace `pcl::uint16_t` -> `std::uint16_t`
 - replace `pcl_isfinite`  -> `std::isfinite`
 - replace `pcl::getFieldIndex(input, name, fields)` -> `pcl::getFieldIndex<PointT>(name, fields)`

### map callback in compare_map
`pcl::PointCloud<pcl::PointXYZ>` type is not supported for subscription. The subscribed topic type is changed to  `sensor_msgs/PointCloud2`, then the received topic is convert to the `pcl::PointCloud<pcl::PointXYZ>` type.

### launch file 
 - preprocessor.launch.py: arguments `input_points_raw0`,`1`,...,`7`, and `input_points_number` are replaced to one `input_points_raw_list` arg since it was difficult to combine multiple args into one string_array. -> needs to change other launch files includes this launch.

### XmlRpc is replaced with string_array
XmlRpc is used to parse the string array from rosparam. In ros2, we can directly get the parameter as a string_array, so the dependency is removed.

### subscriber callback function having more than two args 
Don't use `auto` for the std::bind.
https://answers.ros.org/question/308386/ros2-add-arguments-to-callback/

#### pcl::getPointCloudDifference
`pcl::getPointCloudDifference is deprecated: tgt parameter is not used; it is deprecated and will be removed in future releases`. Should replace? -> replace with the new one (just remove the unused arg).


sensing/preprocessor/pointcloud/pointcloud_preprocessor/src/passthrough_filter/main.cpp
```diff
@@ -0,0 +1,14 @@
...
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  auto node =
+    std::make_shared<pointcloud_preprocessor::PassThroughFilterComponent>(rclcpp::NodeOptions());
+  rclcpp::spin(node);
+  rclcpp::shutdown();
+
+  return 0;
+}
```
mitsudome-r, line 14: Is this main.cpp used anywhere?

sensing/preprocessor/pointcloud/pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -48,269 +27,159 @@ catkin_package(
...
+# ========== Compare Map Filter ==========
+# -- Distance Based Compare Map Filter --
+ament_auto_add_library(distance_based_compare_map_filter SHARED
   src/compare_map_filter/distance_based_compare_map_filter_nodelet.cpp
+  src/filter.cpp
+  )
+target_link_libraries(distance_based_compare_map_filter ${Boost_LIBRARIES})
+rclcpp_components_register_node(distance_based_compare_map_filter
+  PLUGIN "pointcloud_preprocessor::DistanceBasedCompareMapFilterComponent"
+  EXECUTABLE distance_based_compare_map_filter_exe)
```
mitsudome-r, line 44: Since this executable will be used as a replacement of original executable node, I think we should keep the name of the executable to `distance_based_compare_map_filter_node` instead of `distance_based_compare_map_filter_exe`. 
Same with other executables

mitsudome-r (approved):
LGTM, but with few fixes.

sensing/preprocessor/pointcloud/pointcloud_preprocessor/src/passthrough_filter/main.cpp
TakaHoribe: not needed, so I removed.

sensing/preprocessor/pointcloud/pointcloud_preprocessor/CMakeLists.txt
TakaHoribe: change `_exe` to `_node`

mitsudome-r (approved):

PR #120, opened 2020-11-25
Port euclidean cluster

nik-tier4:
Compiles fully with some warnings on my local machine.

The launch files are yet to be ported.


esteve:
@nik-tier4 @mitsudome-r I thought we were going to use the `euclidean_cluster` package from Autoware.Auto

mitsudome-r:
@esteve Yes, that was the reason why we set it to low priority, but the node in AutowareArchitectureProposal also have some features that are not included in Autoware.Auto version (e.g. euclidean cluster in xy-plane) so we wanted to have that feature as Tier IV until similar feature is merged to Autoware.Auto.

PR #121, opened 2020-11-25
Port traffic light visualization

sgermanserrano:
Port `traffic-light-visualization` to ROS2


perception/util/visualizer/traffic_light_visualization/package.xml
```diff
@@ -7,66 +7,18 @@
   <maintainer email="yukihiro.saito@tier4.jp">Yukihiro Saito</maintainer>
   <license>Apache2</license>
 
+  <buildtool_depend>ament_cmake</buildtool_depend>
```
jilaada, line 10: I think you are using `ament_cmake_auto` in your CMakelist so maybe that will need to be reflected here as well.

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
```diff
@@ -18,43 +18,52 @@
...
-  image_transport_.reset(new image_transport::ImageTransport(pnh_));
-
-  pnh_.param<bool>("enable_fine_detection", enable_fine_detection_, false);
+TrafficLightRoiVisualizerNodelet::TrafficLightRoiVisualizerNodelet(const rclcpp::NodeOptions & options)
+: Node("traffic_light_roi_visualizer_node", options)
+{  
+  using std::placeholders::_1;
+  using std::placeholders::_2;
+  using std::placeholders::_3;
+  enable_fine_detection_ = this->declare_parameter("enable_fine_detection", false);
```
jilaada, line 27: You should be able to do this in the initializer list above.

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
```diff
@@ -18,43 +18,52 @@
...
     sync_with_rough_roi_.reset(new SyncWithRoughRoi(SyncPolicyWithRoughRoi(10), image_sub_, roi_sub_,  rough_roi_sub_));
-    sync_with_rough_roi_->registerCallback(boost::bind(&TrafficLightRoiVisualizerNodelet::imageRoughRoiCallback, this, _1, _2, _3));
+    sync_with_rough_roi_->registerCallback(std::bind(&TrafficLightRoiVisualizerNodelet::imageRoughRoiCallback, this, _1, _2, _3));
   } else {
     sync_.reset(new Sync(SyncPolicy(10), image_sub_, roi_sub_));
-    sync_->registerCallback(boost::bind(&TrafficLightRoiVisualizerNodelet::imageRoiCallback, this, _1, _2));
+    sync_->registerCallback(std::bind(&TrafficLightRoiVisualizerNodelet::imageRoiCallback, this, _1, _2));
   }
-  image_transport::SubscriberStatusCallback connect_cb = boost::bind(&TrafficLightRoiVisualizerNodelet::connectCb, this);
-  std::lock_guard<std::mutex> lock(connect_mutex_);
```
jilaada, line 37: What was the rationale behind removing the mutex here and in the method `connectCb()`? Also I noticed that the `connect_mutex_` member variable is still in the header but not used

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
```diff
@@ -18,43 +18,52 @@
...
-  image_pub_ = image_transport_->advertise("output/image", 1, connect_cb, connect_cb);
+  
+  auto timer_callback = std::bind(&TrafficLightRoiVisualizerNodelet::connectCb, this);
+  const auto period_s = 0.1;
+  const auto period_ns =
+    std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::duration<double>(period_s));
+  timer_ = std::make_shared<rclcpp::GenericTimer<decltype(timer_callback)>>(
+    this->get_clock(), period_ns, std::move(timer_callback),
+    this->get_node_base_interface()->get_context());
+  this->get_node_timers_interface()->add_timer(timer_, nullptr);
```
jilaada, line 44: Is this timer needed because SubscriberStatusCallback is not available in ROS2 and/or Foxy?

jilaada (commented):
Thanks for the port here @sgermanserrano! I fetch the changes in this branch and was able to build and run the node using `ros2 launch traffic_light_visualization traffic_light_roi_visualizer.launch.xml`. I just have a couple of questions regarding some changes made with the implementation. I am not be entirely familiar with `image_transport` package and its current state in ROS2 so please forgive my naive questions below.

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
sgermanserrano: indeed, this is done as in #70

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
sgermanserrano: the declaration in the header was missed, I've removed it now. 
I don't think the mutex is needed anymore, if you see in the old code the publisher `image_pub_` needed to use `connect_cb` to `advertise` with `connect_cb` also accessing `image_pub_`, but that is not happening anymore as `image_pub_` is not dependent on `connect_cb`

perception/util/visualizer/traffic_light_visualization/package.xml
sgermanserrano: yeah, fixed now

perception/util/visualizer/traffic_light_visualization/package.xml
```diff
@@ -7,66 +7,18 @@
   <maintainer email="yukihiro.saito@tier4.jp">Yukihiro Saito</maintainer>
   <license>Apache2</license>
```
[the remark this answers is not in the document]
sgermanserrano: Fixed

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
sgermanserrano: In this particular case we only have 1 parameter, but I think we should be consistent across the code base and since in nodes with more parameters we are doing it within the body of the constructor I think it makes more sense to do the same

jilaada: Fair enough.

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
jilaada: I see, thanks for the explanation!

perception/util/visualizer/traffic_light_visualization/src/nodelet.cpp
jilaada: Thanks for linking the existing package!

jilaada (approved):
Thanks for making the changes here. The changes look good to me

PR #123, opened 2020-11-26
add preprocessor.launch.xml launch

TakaHoribe:
This is just an XML wrapper of the `preprocessor.launch.py`, but it might be useful to merge this as an example to load the python launch file from xml.


fred-apex-ai:
@TakaHoribe great, thanks for creating this example!

fred-apex-ai (approved):

PR #125, opened 2020-11-26
Port initial-pose-button-panel

sgermanserrano:
Port `initial-pose-button-panel` to ROS2.

To note:
- As the [client call](https://github.com/tier4/Pilot.Auto/pull/125/files#diff-37aff11e5ae763f660e9df6c6df2d63aeac085d24eeb967d1ce3971493337c44R128) is non-blocking now I've moved the intialization of the button inside the request scope otherwise it'll be initalized before getting a response. 


TakaHoribe (approved):
LGTM. Thank you!

PR #128, opened 2020-11-30
Fix lint issues in pose_initializer

nnmm:


jilaada (approved):
LGTM! I was able to build and and run the tests with no issues

PR #130, opened 2020-11-30
Fix lints in surround_obstacle_checker package

nnmm:
And the issues found by clang-tidy

### DIFF
--- a/planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/CMakeLists.txt
+++ b/planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/CMakeLists.txt
@@ -4,6 +4,8 @@ project(surround_obstacle_checker)
 ### Compile options
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  set(CMAKE_CXX_EXTENSIONS OFF)
 endif()
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
   add_compile_options(-Wno-unused-parameter -Wall -Wextra -Wpedantic)
@@ -12,7 +14,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(eigen3_cmake_module REQUIRED)
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED COMPONENTS common) 
+find_package(PCL REQUIRED COMPONENTS common)
 find_package(ament_cmake_auto REQUIRED)
 ament_auto_find_build_dependencies()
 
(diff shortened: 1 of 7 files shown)


planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/include/surround_obstacle_checker/debug_marker.hpp
```diff
@@ -1,28 +1,28 @@
...
 #include <string>
 
-#include <autoware_planning_msgs/msg/stop_reason_array.hpp>
-#include <geometry_msgs/msg/pose.hpp>
-#include <rclcpp/rclcpp.hpp>
-#include <visualization_msgs/msg/marker.hpp>
-#include <visualization_msgs/msg/marker_array.hpp>
+#include "autoware_planning_msgs/msg/stop_reason_array.hpp"
+#include "geometry_msgs/msg/pose.hpp"
+#include "rclcpp/rclcpp.hpp"
```
sgermanserrano, line 23: @nnmm wouldn't it be better to include system headers using <> instead of "" [https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html](https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html)

nnmm: If you're referring to rclcpp, is that really a system header? It is passed explicitly to the compiler and is not located in any of the system directories.

jilaada (commented):
These changes look good to me. I'll test and approve once the conflicts are resolved. As a side note: I noticed while reviewing different PRs that there are a lot of ways to get the current ROS time in this package `this->clock_->now()` is used. It is a very minor thing but maybe having all of these calls be the same just to ensure that there is no ambiguity regarding the clock type used.

planning/scenario_planning/lane_driving/motion_planning/surround_obstacle_checker/CMakeLists.txt
```diff
@@ -12,7 +14,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(eigen3_cmake_module REQUIRED)
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED COMPONENTS common) 
+find_package(PCL REQUIRED COMPONENTS common)
```
nnmm, line 17: A linter complained about trailing whitespace here

nnmm:
@jilaada The conflicts should be gone now. I'd like having a uniform style for these too – it wouldn't be too hard to bulk-replace it in the code base.

jilaada (commented):
LGTM. I was able to compile and run the node. Though, in order to do without any issues, I needed to add `<depend>lexus_description</depend>` to the `package.xml` so that it would be picked up as a dependency by colcon. Its not really a linter issue but I figured it should be fixed.

nnmm:
@jilaada Good point – one problem is that the vehicle models should be configurable, so you would need a dependency on all of them. It could be better to have a meta-package `vehicle_description` or so that has `exec_depend`s for all the vehicle descriptions. I will do that in a follow-up PR.

jilaada:
> I will do that in a follow-up PR.

Great thanks! So with that I will just merge this PR. Do you need an issue to track the creation of the meta package?

nnmm:
No, I'll just start right away :)

PR #135, opened 2020-12-01
Port object flow fusion

esteve:
This package depends on `eigen_conversions`, which hasn't been ported to ROS 2.

I've filed a ticket upstream about it: https://github.com/ros2/geometry2/issues/350

As a workaround I copied the code from `eigen_conversions` into our codebase. Porting `eigen_conversions` seems straightforward.

Edit: the relevant functions in `eigen_conversions` were added in https://github.com/ros2/geometry2/pull/311, however only in the `ros2` branch, but not in the `foxy` one.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -43,4 +43,20 @@ jobs:
         . /opt/ros/foxy/setup.sh
         colcon test \
           --event-handlers console_cohesion+ \
-          --return-code-on-test-failure
+          --return-code-on-test-failure \
+          --packages-skip \
+            tf2_msgs \
+            tf2_kdl \
+            tf2_py \
+            geometry2 \
+            test_tf2 \
+            tf2_eigen \
+            tf2_ros \
+            tf2_tools \
+            tf2_sensor_msgs \
+            tf2_geometry_msgs \
+            geometry_experimental \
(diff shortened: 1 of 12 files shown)


esteve:
Converting back to a draft, the functions in `eigen_conversions` were ported in https://github.com/ros2/geometry2/pull/311

esteve:
Unfortunately the `tf2_eigen_kdl` package hasn't been released as a Debian package yet, but I'd prefer not to add it to the repos file, it should come out soon.

esteve:
I've added the geometry2 repo temporarily in https://github.com/tier4/Pilot.Auto/pull/135/commits/0457e6063fe9de4da980cd1e4034565b7bac6ffc to see if CI can pass

esteve:
In the end, I've readded the functions from `eigen_conversions` since they have only been added to the `ros2` branch of `geometry2`, but not `foxy`, so they will most likely be released for `galactic`

esteve:
I've submitted https://github.com/ros2/geometry2/pull/356 for review, which adds the functions from `eigen_conversions` to Foxy. I'll update this PR with the changes there and remove the relevant functions here.

esteve:
Update: I managed to do all the conversions using the `tf2::convert` function, so there was no need to port `eigen_conversions` after all.

PR #136, opened 2020-12-01
[gnss_poser] switch broadcaster to full member...

fred-apex-ai:
In porting `sensing_launch`, I started a node with the gnss_poser and ran into 

```
[INFO] [ublox_gps_node-1]: process started with pid [31684]
[ublox_gps_node-1] terminate called after throwing an instance of 'std::runtime_error'
[ublox_gps_node-1]   what():  U-Blox: Could not open serial port :/dev/ttyACM0 open: No such file or directory
[ERROR] [ublox_gps_node-1]: process has died [pid 31684, exit code -6, cmd '/home/frederik.beaujean/AutowareArchitectureProposal/install/ublox_gps/lib/ublox_gps/ublox_gps_node --ros-args -r __node:=ublox -r ublox/fix:=ublox/nav_sat_fix -r __ns:=/sensing/gnss'].
```

This PR avoids this problem by using a proper member instead of a shared pointer because `shared_from_this()` must not be called in the constructor of `this`. It looked like a `shared_ptr` was unnecessary here anyways.

@esteve @mitsudome-r

@jilaada do you remember if you ever started this node when you ported the package?

### DIFF
--- a/sensing/preprocessor/gnss/gnss_poser/CMakeLists.txt
+++ b/sensing/preprocessor/gnss/gnss_poser/CMakeLists.txt
@@ -18,8 +18,8 @@ set(GNSS_POSER_HEADERS
   include/gnss_poser/gnss_stat.hpp
  )
 
-ament_auto_add_library(gnss_poser_node SHARED 
-	src/gnss_poser_core.cpp 
+ament_auto_add_library(gnss_poser_node SHARED
+	src/gnss_poser_core.cpp
 	${GNSS_POSER_HEADERS}
 )
 
--- a/sensing/preprocessor/gnss/gnss_poser/include/gnss_poser/gnss_poser_core.hpp
+++ b/sensing/preprocessor/gnss/gnss_poser/include/gnss_poser/gnss_poser_core.hpp
@@ -67,7 +67,7 @@ private:
 
   tf2::BufferCore tf2_buffer_;
   tf2_ros::TransformListener tf2_listener_;
-  std::shared_ptr<tf2_ros::TransformBroadcaster> tf2_broadcaster_;
+  tf2_ros::TransformBroadcaster tf2_broadcaster_;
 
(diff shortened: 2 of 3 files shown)


jilaada:
@fred-apex-ai Yes, I did start the node, but I guess this was before [I had made the change to `shared_from_this()`](https://github.com/tier4/Pilot.Auto/pull/18#discussion_r506003731).

jilaada:
> @fred-apex-ai Yes, I did start the node, but I guess this was before [I had made the seemingly minor change to `shared_from_this()`](https://github.com/tier4/Pilot.Auto/pull/18#discussion_r506003731). I also gave instructions in the PR description for `gnss_poser` for reviewers to also test the node as well to make sure it wasn't just my system or if I had made any mistakes in building or in my environment for consistency.

jilaada (approved):
I tested this using `ros2 launch gnss_poser gnss_poser.launch.xml` and `ros2 launch gnss_poser ubloxfix2mgrs.launch.xml`and the nodes launch properly without crashing. Thanks for making the fix.

fred-apex-ai:
@jilaada Thanks for merging!

> and the nodes launch properly without crashing.

That's the nature of undefined behavior "works on my machine"

Did you use the same compiler g++9.3.0  inside ADE as I do? cppreference mentions that only from C++17 on, an exception is thrown but my g++ already does with C++14 

https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

jilaada:
> @jilaada Thanks for merging!
> 
> > and the nodes launch properly without crashing.
> 
> That's the nature of undefined behavior "works on my machine"
> 
> Did you use the same compiler g++9.3.0 inside ADE as I do? cppreference mentions that only from C++17 on, an exception is thrown but my g++ already does with C++14
> 
> https://en.cppreference.com/w/cpp/memory/enable_shared_from_this

I had no idea @fred-apex-ai, thanks for the explanation! I think at this point i may have been using an environment outside of ADE as this was the first package I ported. Definitely something to take note of for next time!

PR #144, opened 2020-12-03
Use quotes for includes where appropriate

nnmm:
Make all includes, with the exception of system (standard library) C and C++ header files, use quotes. This is so that the `ament_cpplint` linter doesn't get confused.


awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.hpp
```diff
@@ -12,10 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <rclcpp/rclcpp.hpp>
+#include "rclcpp/rclcpp.hpp"
```
esteve, line 15: IMHO this and the rest should be angle brackets, as it's a system dependency. What is the error that cpplint complains about? The order of the includes?

esteve (changes requested):

awapi/awapi_awiv_adapter/include/awapi_awiv_adapter/awapi_autoware_state_publisher.hpp
nnmm: Yes, and it just calls everything that's in angle brackets and has a file extension a "C system dependency"

nnmm: What is the definition of a system dependency though? As I read this, rclcpp is not a system dependency: https://github.com/tier4/Pilot.Auto/pull/130

nnmm: Nevermind, that link doesn't define it. Anyway, the Google Style guide uses quotes for (non-standard) library includes as well in an example.

sgermanserrano: This [link](https://docs.microsoft.com/en-us/cpp/preprocessor/hash-include-directive-c-cpp?view=msvc-160) contains a more detailed explanation of the search order for the includes

nnmm: Yes, both styles compile fine because those headers' locations are passed by `-I` although the differences in search path behavior might make angle brackets slightly more robust here. But like you say that's not the ROS convention: As you can see [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/topic_statistics/subscription_topic_statistics.hpp), the ROS sources use quotes for files from the same package as well as other packages/libraries.

So to recap:
* Quotes are the convention for includes in ROS, except for C and C++ standard library headers.
* `ament_cpplint` also assumes that only C and C++ standard library headers (which it calls "system headers") use angle brackets. E.g. `<awapi_awiv_adapter/awapi_autoware_util.hpp>` would be considered a C system header, because it has an extension.
* `ament_cpplint` imposes the order "corresponding header file, c system, c++ system, other" – i.e. "other" includes all headers from the same package, other packages, or third-party libraries.
* I think this ordering is suboptimal, see [this article](https://blog.knatten.org/2010/07/01/the-order-of-include-directives-matter/). Missing includes in headers from AutowareAuto/Pilot.Auto packages have a higher chance of revealing themselves when including C++ system headers afterwards.
* So we can either discuss this in the `ament_cpplint` repo, **or let the two wrongs cancel out**: I.e. keep angle brackets and let `ament_cpplint` believe that `<rawapi_awiv_adapter/awapi_autoware_util.hpp>` is a C system header, and use the better ordering "corresponding header file, AutowareAuto/Pilot.Auto headers, c++ system, ROS and third-party headers". What do you think @esteve and others?

esteve: I tend to use angle brackets for packages from outside of the current package (e.g. rclcpp, string, etc.) and quotes for headers that are from the current package. But after checking some other packages (e.g. demo_nodes_cpp), I agree that it's fine to use quotes everywhere.

esteve (approved):

esteve:
@nnmm I've changed my review to approve, let me know if you'd like to merge this or you have more changes.

nnmm:
@esteve Ok, then let's merge this one and https://github.com/tier4/scenario_runner.iv.universe/pull/37#pullrequestreview-545737112

PR #146, opened 2020-12-07
Enable lints in astar_search

nnmm:
clang-tidy also pointed out that the `fmod` function is not `constexpr`, so the function will never return a `constexpr` value.

### DIFF
--- a/planning/scenario_planning/parking/astar_search/src/astar_search.cpp
+++ b/planning/scenario_planning/parking/astar_search/src/astar_search.cpp
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "astar_search/astar_search.hpp"
+
 #include "astar_search/helper.hpp"
 
 #include "tf2/utils.h"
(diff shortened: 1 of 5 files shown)


planning/scenario_planning/parking/astar_search/src/astar_search.cpp
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include "astar_search/astar_search.hpp"
+
```
jilaada, line 16: Curious, why was a new line added here?

jilaada (approved):
I was able to compile this packages and the tests ran without errors. Changes look good to me. Just a small question about the addition of a new line below:

planning/scenario_planning/parking/astar_search/src/astar_search.cpp
nnmm: As the "corresponding header file", I thought it should be in its own group. Cpplint thinks so too: If there were any C++ stdlib headers, they would be put after this header, but before the helper header by cpplint.

jilaada: i see, thanks for the explanation!

PR #147, opened 2020-12-07
Enable lints in freespace_planner

nnmm:
Nothing from clang-tidy.


jilaada (approved):
I was able to build and run the following package. I also ran clang tidy on the source files of this package to verify. Changes look good to me.

PR #151, opened 2020-12-08
Run uncrustify on the entire Pilot.Auto codebase

nnmm:
This is to make future linter PRs smaller.


sgermanserrano:
it'll be worth waiting for the current open PRs that are porting nodes to be merged before this as it is likely to create conflicts

nnmm:
@sgermanserrano Yeah, it's probably a bit of a hassle for those PRs. I removed those packages from this PR, hope I got them all. I think waiting for them would take too long and there would be new PRs by then.

@sgermanserrano @nik-tier4 @esteve Ideally, you'd follow up your PRs with an uncrustification PR, but if not, that will happen when we turn on linters anyway.

sgermanserrano:
I think most of the open PRs have already been reviewed and just waiting to be merged

esteve (approved):

PR #152, opened 2020-12-08
Add metapackage for all vehicle descriptions

nnmm:
The motivation is that the vehicle models used by launch files should be configurable, so the packages with these launch files need a dependency on all possible vehicle models that exist. Currently there is only one vehicle description, but I think Tier IV has more internally and AutowareAuto might add others in the future too.

Instead of adding a list of description packages to each package that uses a vehicle model, which would need to be changed when new models are added, this PR groups the vehicle descriptions into one package that can be used as a dependency that won't change.

This was modeled after the ament_lint_common metapackage.

## Issues
Maybe this would be better placed next to `lexus_description`, but `lexus_description` is its own repository. Maybe it can be renamed to `vehicle_description` and then include `lexus_description` and `all_vehicle_descriptions` as subdirectories.

I am happy to accept suggestions for a better name.

### DIFF
--- a/build_depends.repos
+++ b/build_depends.repos
@@ -15,3 +15,8 @@ repositories:
     type: git
     url: https://github.com/ApexAI/topic_tools.git
     version: autoware
+  vendor/lexus_description:
+    type: git
+    url: https://github.com/tier4/lexus_description.iv.universe.git
+    version: ros2
+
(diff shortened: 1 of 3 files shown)


jilaada:
I think `vehicle_descriptions` would be a good meta-package name. Each sub-package should be prefixed with the vehicle type/model that the description it contains describes.

nnmm:
@mitsudome-r What do you think about renaming & generalizing the `lexus_description` repository?

nnmm:
Related: I'm also opening PRs to make launch files configurable, e.g. https://github.com/tier4/Pilot.Auto/pull/228, and to pipe that parameter through the higher-level launch files: https://github.com/tier4/autoware_launcher.iv.universe/pull/20.

PR #154, opened 2020-12-09
Added linters to obstacle_stop_planner

esteve:


jilaada (changes requested):
I'm getting quite a number of warning from the compiler (before running tests). Should these be fixed in this PR? Tests pass though so maybe its not an issue? Also `clang-tidy` gave an error when checking `node.cpp`

esteve:
@jilaada not sure about we should do with the warnings, but I'm fixing the clang-tidy errors, thanks for noticing.

jilaada:
@kenji-miyake Do you know if these warnings have been fixed upstream? If this isn't one of the packages you have worked on, do you know who might be able to let us know? 

@esteve Thanks for addressing the `clang-tidy` error

kenji-miyake:
@jilaada No, we didn't focus on these issues.
Of course, we hope we could fix warnings, so I think this PR is nice! Thank you!

esteve:
@jilaada should I fix the warnings here, wait for upstream or just merge it as is?

jilaada:
> @jilaada should I fix the warnings here, wait for upstream or just merge it as is?

It seems like the warnings were not fixed upstream so maybe if they were fixed in this PR it would be good - I don't think the changes are that significant.

esteve:
@jilaada ok, I'll fix them here.

esteve:
@jilaada warnings fixed, could you have a look at this again? Thanks.

jilaada (approved):
@esteve There were a couple of additional type mismatches which I fixed in d08575a. Will merge once CI passes

esteve:
@jilaada thanks for making the changes 🙂

PR #156, opened 2020-12-09
Add linters to mission_planner

esteve:


planning/mission_planning/mission_planner/package.xml
```diff
@@ -17,6 +17,9 @@
   <depend>tf2_ros</depend>
   <depend>tf2_geometry_msgs</depend>
 
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
```
jilaada, line 21: I think we can change this to `ament_cmake_cppcheck` in case the linter checks fail due to anything other than cppcheck

jilaada (changes requested):
It seems like the `CMakeLists.txt` is missing the linter tests?

planning/mission_planning/mission_planner/package.xml
esteve: @jilaada I've removed `ament_lint_common` and added `ament_cmake_cppcheck` in c101295b. I fixed the linting errors when the rest of the linters were enabled, should I revert those changes and only fix those that `cppcheck` detects?

jilaada: If you've already made those fixes it's a bonus so its better to just leave them. Thanks for changing it to `ament_cmake_cppcheck` though!

planning/mission_planning/mission_planner/CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ project(mission_planner)
 ### Compile options
 if(NOT CMAKE_CXX_STANDARD)
   set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  set(CMAKE_CXX_EXTENSIONS OFF)
```
jilaada, line 8: It looks like the linter tests haven't been added to the CMakelist yet?

esteve: Fixed in bca948a0de2c2e0faebfcdfe2bd695ffbf69ae93

jilaada (approved):
LGTM!

PR #165, opened 2020-12-10
ROS2 Linting: geo_pos_conv

jilaada:
## Summary

Small and straightforward lint of the `geo_pos_conv` package. I converted the `CMakelist.txt` to use `ament_cmake_auto`. 

Notes to Reviewer:
* Most of the changes here is converting `(double)` to `static_cast<double>()`. Please ensure that brackets are placed correctly.

## Testing

1.  Compile with the correct build flags
```
colcon build --packages-up-to geo_pos_conv --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=1
```

2. Run the linter tests
```
colcon test --packages-select geo_pos_conv && colcon test-result --verbose
```

3. Run `clang-tidy` on the the source files from the root `AutowareArchitectureProposal` directory
```
clang-tidy -p build/compile_commands.json src/autoware/autoware.iv/sensing/preprocessor/gnss/geo_pos_conv/include/geo_pos_conv/geo_pos_conv.hpp
clang-tidy -p build/compile_commands.json src/autoware/autoware.iv/sensing/preprocessor/gnss/geo_pos_conv/src/geo_pos_conv.cpp
```

### DIFF
--- a/sensing/preprocessor/gnss/geo_pos_conv/src/geo_pos_conv.cpp
+++ b/sensing/preprocessor/gnss/geo_pos_conv/src/geo_pos_conv.cpp
@@ -145,11 +145,16 @@ void geo_pos_conv::set_plane(int num)
     lon_min = 0;
     lat_deg = 154;
     lat_min = 0;
+  } else {
+    lon_deg = 0;
+    lon_min = 0;
+    lat_deg = 0;
+    lat_min = 0;
   }
 
   // swap longitude and latitude
-  m_PLo = M_PI * ((double)lat_deg + (double)lat_min / 60.0) / 180.0;
-  m_PLato = M_PI * ((double)lon_deg + (double)lon_min / 60.0) / 180;
+  m_PLo = M_PI * static_cast<double>(lat_deg + static_cast<double>(lat_min / 60.0) / 180.0);
+  m_PLato = M_PI * static_cast<double>(lon_deg + static_cast<double>(lon_min / 60.0) / 180);
 }
 
 void geo_pos_conv::set_xyz(double cx, double cy, double cz)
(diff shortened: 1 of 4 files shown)


sensing/preprocessor/gnss/geo_pos_conv/src/geo_pos_conv.cpp
```diff
@@ -209,68 +214,75 @@ void geo_pos_conv::conv_llh2xyz(void)
   AW = 6378137.0;            // Semimajor Axis
   FW = 1.0 / 298.257222101;  // 298.257223563 //Geometrical flattening
 
-  Pe = (double)sqrt(2.0 * FW - pow(FW, 2));
-  Pet = (double)sqrt(pow(Pe, 2) / (1.0 - pow(Pe, 2)));
+  Pe = static_cast<double>(sqrt(2.0 * FW - pow(FW, 2)));
+  Pet = static_cast<double>(sqrt(pow(Pe, 2) / (1.0 - pow(Pe, 2))));
 
-  PA = (double)1.0 + 3.0 / 4.0 * pow(Pe, 2) + 45.0 / 64.0 * pow(Pe, 4) +
```
nnmm, line 215: I think these casts were unnecessary. First, the casts have higher precedence than arithmetic operators, so e.g. here it only applied to the 1.0 literal which is already a double. In the cases where it is applied to e.g. `sqrt`, it would be casted to double anyway because the result is stored in a double. Therefore please remove all static casts in this file as well.

jilaada:
@nnmm I decided to throw an error instead of setting the default lateral and longitudinal values to zero since I don't think it makes sense and could produce errors downstream that may not be easily detected.

nnmm (approved):

PR #167, opened 2020-12-10
Added linters to shift_decider

esteve:
Only cpplint and cppcheck


jilaada:
@esteve I think the linter tests are failing here.

esteve:
@jilaada thanks for looking into, they should be fixed now.

jilaada (approved):
LGTM, ran the linter tests and clang-tidy all of which reported no errors

PR #174, opened 2020-12-11
adding linters to shape_estimation

nik-tier4:
Some variables where uninitialized and were giving warnings. Initialized them to zero.
Not sure if it is correct. 

Also removed variables that were not used.


jilaada (approved):
Changes make sense. Tested with clang-tidy (although running it on such a large package was very tedious) and ran the linter tests. LGTM :+1:

Issue #129, opened 2020-11-30
Cmake warning in building pointcloud_preprocessor


fred-apex-ai:
`pointcloud_preprocessor` builds but issues a warning. 

```
--- stderr: pointcloud_preprocessor 
CMake Warning (dev) at CMakeLists.txt:18 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  CMake variable PCL_ROOT is set to:

    /usr

  For compatibility, CMake is ignoring the variable.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

There are other packages that rely on `PCL` and don't have this warning, so we should be able to get rid of it

jilaada:
@fred-apex-ai During the linting of pointcloud_preprocessor I thought I would just try to resolve this issue as well, please take a look: https://github.com/tier4/Pilot.Auto/pull/180

jilaada:
I think we can close this issue now that #180 is merged with this fixes.

PR #180, opened 2020-12-11
ROS2 Linting: pointcloud_preprocessor

jilaada:
## Summary

This looks like a lot of changes but really this is the result of a file being renamed and git not displaying the diff properly. There were a file which had a `.h` extension which I converted to a `.hpp` though another file already had that name so that file name needed to be changed as well.
1. `passthrough_uint16.h` -> `passthrough_uint16.hpp`
2. `passthrough_uint16.hpp` -> `passthrough_uint16_impl.hpp`

Let me know if you think the names don't fit as I don't think its the correct use of "impl" but the `#ifndef` tag seems to suggest that is the case.


sensing/preprocessor/pointcloud/pointcloud_preprocessor/package.xml
```diff
@@ -30,6 +30,9 @@
   <depend>libopencv-dev</depend>
   <depend>libpcl-all-dev</depend>
 
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_cmake_cppcheck</test_depend>
```
nnmm, line 34: Esteve let me know that this should be`ament_cppcheck`, without the `cmake`.

nnmm (commented):
When I batch renamed files from *.h to *.hpp, this was skipped because the hpp file existed already. Now I've taken a second look, and I don't see why this needs to be in two separate files. It is one class template, but two methods are defined in the *_impl.hpp. Maybe it would be best to just merge them into `passthrough_uint16.hpp`?

sensing/preprocessor/pointcloud/pointcloud_preprocessor/package.xml
nnmm: Just saw your comment on the other PR – in that case nevermind, `ament_cmake_cppcheck` seemed the right option to me anyway, since that is the one included by `ament_lint_common`.

jilaada:
> When I batch renamed files from *.h to *.hpp, this was skipped because the hpp file existed already. Now I've taken a second look, and I don't see why this needs to be in two separate files. It is one class template, but two methods are defined in the *_impl.hpp. Maybe it would be best to just merge them into `passthrough_uint16.hpp`?

@nnmm Thanks for the comments. So I removed the `_impl` file and merged the two files together. Now the file changes should look like this:
1. `passthrough_uint16.h` -> merged into `passthrough_uint16.hpp`
2. `passthrough_uint16.hpp` -> no change

I also made some CMake changes to get rid of compiler warnings, see #129

sensing/preprocessor/pointcloud/pointcloud_preprocessor/CMakeLists.txt
```diff
@@ -15,8 +17,6 @@ ament_auto_find_build_dependencies()
 find_package(OpenCV REQUIRED)
 find_package(Eigen3 REQUIRED)
 find_package(Boost REQUIRED)
-find_package(PCL REQUIRED)
```
fred-apex-ai, line 18: @jilaada was removing that line enough to fix #129 ?

jilaada: Yeah I checked the error message: https://cmake.org/cmake/help/git-stage/policy/CMP0074.html and it seemed to relate to the find_package and the use of ament_auto to find dependencies. Removing these 2 lines did the trick. I also noticed that the libraries might not be linked correctly, so I added those down below. I'm not a CMake expert so any critique is more than welcomed!

fred-apex-ai: One of the quirks of `ament_auto_find_build_dependencies()` is that it doesn't add the `REQUIRED` keyword, so if you want to require dependencies, you have switch the order. That might have caused the problem, so try

```cmake
find_package(Boost REQUIRED)
find_package(PCL REQUIRED)
ament_auto_find_build_dependencies()
```

jilaada: So I just tried this and I ended up with stderr outputs:
```
jilada@ade:~/AutowareArchitectureProposal (ros2 $% u=)$ colcon build --packages-up-to pointcloud_preprocessor --cmake-args -DCMAKE_EXPORT_COMPILE_COMMANDS=1
Starting >>> pcl_conversions
Finished <<< pcl_conversions [0.21s]
Starting >>> pcl_ros
Finished <<< pcl_ros [0.18s]
Starting >>> pointcloud_preprocessor
[Processing: pointcloud_preprocessor]
--- stderr: pointcloud_preprocessor
** WARNING ** io features related to openni will be disabled
** WARNING ** io features related to openni2 will be disabled
** WARNING ** io features related to pcap will be disabled
** WARNING ** io features related to png will be disabled
** WARNING ** io features related to libusb-1.0 will be disabled
** WARNING ** visualization features related to openni will be disabled
** WARNING ** visualization features related to openni2 will be disabled
** WARNING ** apps features related to openni will be disabled
---
Finished <<< pointcloud_preprocessor [51.6s]

Summary: 3 packages finished [52.5s]
  1 package had stderr output: pointcloud_preprocessor
```

nnmm: Yep, many packages have that. I didn't find out how to suppress them yet.

nnmm (approved):
I'd re-add PCL like Fred suggested, otherwise looks good to me

jilaada:
@fred-apex-ai @nnmm I have updated the CMake as suggested, see 2877403

PR #209, opened 2020-12-15
Fix ekf var initialization & launch

TakaHoribe:
ekf_localiser: fix initialization & launch


TakaHoribe:
confirmed in experiment

PR #212, opened 2020-12-15
Replace waitForTransform by lookupTransform in surround_obstacle_checker

nnmm:
Part of https://github.com/tier4/AutowareArchitectureProposal/issues/132
The result of the waitForTransform doesn't even seem to have been used, I fixed it and moved the variable declarations to their initialization.


mitsudome-r (approved):
LGTM

PR #213, opened 2020-12-15
Replace waitForTransform with lookupTransform in multi_object_tracker

nnmm:
Closer to the ROS1 original now: https://github.com/tier4/Pilot.Auto/blob/master/perception/object_recognition/tracking/multi_object_tracker/src/node.cpp#L39-L59


nnmm:
Part of https://github.com/tier4/AutowareArchitectureProposal/issues/132

mitsudome-r (approved):

PR #228, opened 2020-12-16
Make launch files load vehicle model params dependent on argument

nnmm:
Make the vehicle model configurable in all launch files that require vehicle information.


mitsudome-r (approved):
LGTM

PR #238, opened 2020-12-23
modify launch file for making psim work

k0suke-murakami:
## What is this pr
* fix minor issues in launch files
* need to be merged for making planning simulator work

### DIFF
--- a/simulator/util/dummy_perception_publisher/launch/dummy_perception_publisher.launch.xml
+++ b/simulator/util/dummy_perception_publisher/launch/dummy_perception_publisher.launch.xml
@@ -40,20 +40,20 @@
   </group>
 
   <!-- preception module -->
-  <group ns="perception">
+  <group>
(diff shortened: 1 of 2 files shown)


simulator/util/dummy_perception_publisher/launch/dummy_perception_publisher.launch.xml
```diff
@@ -40,20 +40,20 @@
   </group>
 
   <!-- preception module -->
-  <group>
+  <group ns="perception">
```
mitsudome-r, line 43: We cannot use `ns` with `group`. Instead, we are using `push-ros-namespace` which is there already.

k0suke-murakami: @mitsudome-r  Thank you for the comment! I commit to removing unnecessary 'ns'

mitsudome-r (approved):
LGTM

PR #246, opened 2021-01-07
Fix bugs in autoware vehicle rviz plugin

wep21:
### How to test
Please check if you can see vehicle plugins at the upper left on rviz as below.
Screenshot from 2021-01-08 06-21-52


TakaHoribe (approved):
LGTM

PR #250, opened 2021-01-10
traffic_light_classifier: Fix engine save dir

wep21:
### How to test
- launch cnn classfier
`ros2 launch traffic_light_classifier traffic_light_classifier.launch.xml use_gpu:=true`
- check if engine file is in the package share directory
`ls install/traffic_light_classifier/share/traffic_light_classifier/data/traffic_light_classifier_mobilenetv2.engine`


k0suke-murakami (approved):
LGTM

PR #296, opened 2021-01-28
[gnss_poser] fix link to Geographiclib

mitsudome-r:
This fixes missing link to Geographic lib package.
It has been tested with an actual sensor data.


wep21 (approved):
LGTM

PR #306, opened 2021-02-02
Ros2 v0.8.0 object range splitter

wep21:


k0suke-murakami:
- [x]  Build passed
 Package: object range splitter
- [x] ros2 run success.
 Node : object range splitter_node
- [x] Test  passed

PR #331, opened 2021-02-12
Add missing commits from v0.7

mitsudome-r:
There were mistakes when master was upgraded to version 0.7.
This adds missing commits for v0.7 update.

## How to review
Make sure that the repository can be built with instructions in https://github.com/tier4/AutowareArchitectureProposal.proj


UMiho (approved):
Thank you for submitting this PR. I leave comments which we talked offline:
- This is a correction for the case where missed commits were found between v0.5.0 and v0.7.0.
- Each review of commits is completed in the original private repository, and I skipped it here.

PR #337, opened 2021-02-16
Ros2 v0.8 path_test_publisher

isamu-takagi:


planning/path_test_publisher/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<package format="2">
```
wep21, line 2: It's better to use format 3.

wep21:
test result
``` 
ros2 topic echo /path
header:
  stamp:
    sec: 1613468320
    nanosec: 724778518
  frame_id: map
points:
- pose:
    position:
      x: 0.0
      y: 0.0
      z: 0.0
    orientation:
      x: 0.0
      y: 0.0
      z: 0.0
      w: 1.0
  twist:
    linear:
      x: 10.0
      y: 0.0
      z: 0.0
    angular:
      x: 0.0
      y: 0.0
      z: 0.0
  type: 0
```

planning/path_test_publisher/package.xml
isamu-takagi: Fixed.

wep21 (approved):
LGTM

PR #340, opened 2021-02-16
[mpc follower]: fix timesource

wep21:
Resolve std::runtime_error below.
```
[mpc_follower-27] terminate called after throwing an instance of 'std::runtime_error'
[mpc_follower-27]   what():  can't subtract times with different time sources [1 != 2]
```


TakaHoribe (approved):
LGTM

PR #362, opened 2021-02-25
Fix autoware error monitor

kenji-miyake:
Changes:
- d31a9c2
    - This diag is not output anymore.
- 0c9757a
    - network_usage and ntp_offset cause errors if we don't specify appropriate device names.
    - There is no problem if we ignore this in planning_simulator.
- fb0d251
    - Remove unnecessary parameters. 
        - multiple `path: autoware`
    - Set log-level
    - Apply format
- 973d3bb
    - Apply format

The test result is here. No emergency happens after setting a route.
image


tkimura4 (approved):
LGTM!

PR #369, opened 2021-02-26
add nullptr check when publish concatenate data

taichiH:
I fixed concatenate_data bug.

I've got this error when input nullptr as a publish message at here https://github.com/tier4/AutowareArchitectureProposal.iv/blob/9a32729ef7d3d8bdbd9ab4189184aab58d78d093/sensing/preprocessor/pointcloud/pointcloud_preprocessor/src/concatenate_data/concatenate_data_nodelet.cpp#L269

```
[component_container-3] terminate called after throwing an instance of 'rclcpp::exceptions::RCLInvalidArgument'
[component_container-3]   what():  failed to publish message: ros_message argument is null, at /tmp/binarydeb/ros-foxy-rcl-1.1.10/src/rcl/publisher.c:317`
```


kenji-miyake (approved):
I approve because it's correct, but I'd like to know why it happens. There isn't nullptr check in our ROS1 code as well. :thinking: 
@taichiH Do you have any idea?

PR #397, opened 2021-03-02
fix max velocity visualization

tkimura4:
Fix the bug that the max velocityis not visualized in rviz.

image


mitsudome-r:
I've changed the target branch to ros2.

wep21 (approved):
LGTM

PR #409, opened 2021-03-08
fix obstacle_stop_planner

tkimura4:
Fix implement miss in obstacle_stop_planner.
https://github.com/tier4/AutowareArchitectureProposal.iv/blob/master/planning/scenario_planning/lane_driving/motion_planning/obstacle_stop_planner/src/node.cpp#L193

(At the current code, the trajectory is not extended regardless of the value of stop_planner.extend_distance)


k0suke-murakami (approved):
LGTM

PR #415, opened 2021-03-10
add timeout_ignore_state to emergency_handler

tkimura4:
Add timeout_ignore_state to emergency_handler.

## Modification
When _is_in_timeout_ignore_state_ is True (autoware/state is INITIALIZING_VEHICLE), _hazard_status.level_ does not become ERROR by timeout of topics such as driving_capability.
The reason for the above change is that frequencies of these topics become unstable soon after starting Autoware.

### DIFF
--- a/system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
+++ b/system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
@@ -383,7 +383,7 @@ autoware_system_msgs::msg::HazardStatus EmergencyHandler::judgeHazardStatus()
     const auto is_in_heartbeat_timeout_ignore_state =
       (autoware_state_->state == AutowareState::INITIALIZING_VEHICLE);
 
-    if (heartbeat_driving_capability_->isTimeout() && !is_in_heartbeat_timeout_ignore_state) {
+    if (!is_in_heartbeat_timeout_ignore_state && heartbeat_driving_capability_->isTimeout()) {
       RCLCPP_WARN_THROTTLE(
         this->get_logger(), *this->get_clock(), std::chrono::milliseconds(1000).count(),
         "heartbeat_driving_capability is timeout");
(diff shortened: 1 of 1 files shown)


system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
```diff
@@ -407,15 +405,14 @@ autoware_system_msgs::msg::HazardStatus EmergencyHandler::judgeHazardStatus()
           "heartbeat_is_state_timeout is timeout"));
     }
 
-    if (is_state_timeout_->is_timeout) {
+    if (is_state_timeout_->is_timeout && !is_in_timeout_ignore_state) {
```
kenji-miyake, line 408: I think we should ignore only heartbeat timeouts. This is not a heartbeat.
So please rename `is_in_timeout_ignore_state` to `is_in_heartbeat_timeout_ignore_state` and don't ignore is_state_timeout_->is_timeout.

tkimura4: Dealt with it @ 7b2aee5

system/emergency_handler/src/emergency_handler/emergency_handler_core.cpp
```diff
@@ -396,7 +394,7 @@ autoware_system_msgs::msg::HazardStatus EmergencyHandler::judgeHazardStatus()
           "heartbeat_driving_capability is timeout"));
     }
 
-    if (heartbeat_is_state_timeout_->isTimeout()) {
+    if (heartbeat_is_state_timeout_->isTimeout() && !is_in_heartbeat_timeout_ignore_state) {
```
kenji-miyake, line 397: and one more small thing, writing `is_in_heartbeat_timeout_ignore_state` before `isTimeout()` seems better.

```cpp
    if (!is_in_heartbeat_timeout_ignore_state && heartbeat_is_state_timeout_->isTimeout()) {
```

kenji-miyake (approved):

PR #423, opened 2021-03-12
Fix/ll2 marker

taikitanaka3:
## PRの種類

- [x] bugfix Rviz error at no point marker

check if rviz doesn't show error using this map 
image

https://drive.google.com/drive/folders/1oKt9bZrd6AQSPGyBY4oaSwU3wtgSlhV4?usp=sharing


mitsudome-r (approved):
LGTM

PR #429, opened 2021-03-15
Add depend tag for image_transport_plugins

isamu-takagi:
Fix runtime error: `terminate called after throwing an instance of 'image_transport::TransportLoadException'`


wep21:
Why do you add dependency into `lane_departue_checker`?
I guess this module does not use image.

isamu-takagi:
@wep21 I made a mistake when editing. Fixed it with this PR.
https://github.com/tier4/autoware_launcher.iv.universe/pull/95

PR #434, opened 2021-03-16
Fix typo in system module

kmiya:
Fix typo in system module.

@ito-san 
The followings are just my humble suggestion. Do you have any other good words to replace them?
- entried -> appended
- spinned -> spin()

### DIFF
--- a/system/system_monitor/include/system_monitor/utils.hpp
+++ b/system/system_monitor/include/system_monitor/utils.hpp
@@ -29,7 +29,7 @@
 /// \tparam MonitorT Monitor type
 /// \param monitor_ptr Shared pointer of a monitor node to be spin().
 /// \param period Spin period.
-template <typename MonitorT>
+template<typename MonitorT>
 void spin_and_update(const std::shared_ptr<MonitorT> & monitor_ptr, std::chrono::seconds period)
 {
   while (rclcpp::ok()) {
--- a/system/system_monitor/include/system_monitor/utils.hpp
+++ b/system/system_monitor/include/system_monitor/utils.hpp
@@ -19,7 +19,7 @@
 
 /// This function will spin the given node and update the diagnostic state at each iteration
 /// \tparam MonitorT Monitor type
-/// \param monitor_ptr Shared pointer of a monitor node to be spinned.
+/// \param monitor_ptr Shared pointer of a monitor node to be spin().
 /// \param period Spin period.
 
 #ifndef SYSTEM_MONITOR__UTILS_HPP_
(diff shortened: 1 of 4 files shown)


system/system_monitor/src/net_monitor/net_monitor.cpp
```diff
@@ -124,7 +124,7 @@ void NetMonitor::checkUsage(diagnostic_updater::DiagnosticStatusWrapper & stat)
 
       ++index;
       close(fd);
-      entried.push_back(ifa->ifa_name);
+      appended.push_back(ifa->ifa_name);
```
kenji-miyake, line 127: I think `interface_names` is better.

system/system_monitor/include/system_monitor/utils.hpp
```diff
@@ -19,7 +19,7 @@
 
 /// This function will spin the given node and update the diagnostic state at each iteration
 /// \tparam MonitorT Monitor type
-/// \param monitor_ptr Shared pointer of a monitor node to be spinned.
+/// \param monitor_ptr Shared pointer of a monitor node to be spin().
 /// \param period Spin period.
```
kenji-miyake, line 23: How about moving this right before the function (around Line32)?

kenji-miyake: Also, using `spin() in already spinning node` is not allowed in ROS2, so we should confirm whether we can still use this function or not.

ito-san (approved):
LGTM other than comments made by @kenji-miyake.

system/system_monitor/src/net_monitor/net_monitor.cpp
kmiya: Changed in f06568c1dc43c07e8b54747d0c09b02bf0a088c5

system/system_monitor/include/system_monitor/utils.hpp
kmiya: Moved comments in 3213d564af40f917e23acc3ab6108b66c4afe77d (and slightly formatted).
Deleting the function might be out-of-scope of this PR. Better to do on another PR?

kenji-miyake: @kmiya Thank you!
> Deleting the function might be out-of-scope of this PR. Better to do on another PR?

Yes, it's fine.

kenji-miyake:
@kmiya Could you confirm the failure reason of CI and fix it?

kmiya:
Tests have passed :tada:

kenji-miyake (approved):

PR #435, opened 2021-03-16
Fix typo in vehicle module

kmiya:
Fix typo in the vehicle module. `/vehcle/engage` seems to be problematic.


kenji-miyake (approved):
LGTM.

PR #442, opened 2021-03-16
Ros2 fix ndt align srv

isamu-takagi:
Fix this runtime error.
```
[ndt_scan_matcher-34] terminate called after throwing an instance of 'std::runtime_error'
[ndt_scan_matcher-34]   what():  No InputSource
```


kenji-miyake (approved):
@taichiH has tested this on a real vehicle.

PR #461, opened 2021-03-29
Rviz overlay render fix

adamdbrw:
Restores MFD superclass for SteeringAngle plugin. 

Moves rendering to update() function called from the Display superclass in the rendering thread. This fixes the issue.

Note that other plugins can be similarly modified.

### DIFF
--- a/common/util/rviz_plugins/autoware_vehicle_rviz_plugin/src/tools/steering_angle.cpp
+++ b/common/util/rviz_plugins/autoware_vehicle_rviz_plugin/src/tools/steering_angle.cpp
@@ -102,21 +102,7 @@ void SteeringAngleDisplay::onInitialize()
 
   overlay_->show();
 
-  overlay_->updateTextureSize(property_length_->getInt(), property_length_->getInt());
-  overlay_->setPosition(property_left_->getInt(), property_top_->getInt());
-  overlay_->setDimensions(overlay_->getTextureWidth(), overlay_->getTextureHeight());
-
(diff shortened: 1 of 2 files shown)


common/util/rviz_plugins/autoware_vehicle_rviz_plugin/src/tools/steering_angle.cpp
```diff
@@ -94,29 +94,15 @@ SteeringAngleDisplay::~SteeringAngleDisplay()
 
 void SteeringAngleDisplay::onInitialize()
 {
-  RTDClass::onInitialize();
+  MFDClass::onInitialize();
```
mitsudome-r, line 97: I think I said that we can switch back to MFDClass in the previous meeting, but after having second thought, I think we can keep it as RTDClass. The change on moving visualization process to `update` function from `processMessage` can still be applied for stability and process time.
I apologize for going back and forth, but could you make the super class to RTD?

mitsudome-r (commented):
Thanks for the review. I have made a comment about the super class.

common/util/rviz_plugins/autoware_vehicle_rviz_plugin/src/tools/steering_angle.cpp
adamdbrw: in all the plugins?

mitsudome-r (approved):
LGTM

PR #463, opened 2021-03-29
fix: re-construct laneletMapLayer for reindex RTree

vios-fish:
## Kind of PR
- [x] Bugfix

## Abstruction
AutowareOsmParser is changed point value after RTree is constructed. https://github.com/tier4/AutowareArchitectureProposal.iv/blob/2328f517117a2f5f540ee5080a48748a1c2aa60f/map/lanelet2_extension/lib/autoware_osm_parser.cpp#L38-L45
For reindex RTree, construct LaneletMap class again after overwrite points as local one.


mitsudome-r (approved):
LGTM

PR #466, opened 2021-04-02
Fix issues in hdd_reader

kmiya:
## What kinds of PR

- [X] Bug fix

## What is this PR

- Update the launch command
- Fix some issues detected by Coverity Scan
  - Resource leak (`new_sock`)
  - Uninitialized variables
  - Improper null termination (`buf`)
    - Ref. [CWE-170](https://cwe.mitre.org/data/definitions/170.html)
    - Almost all cases seem to be no problem, but just in case fixed it
- Fix some issues detected by Clang-Tidy
  - Deprecated headers
  - Use `using` instead of `typedef`
  - Redundant `void`
  - Use range-based `for`
  - Unused namespace alias
  - Use `nullptr` instead of `0` to initialize pointers

Do __not__ fix the other issues, for example:
- Use `std::array<>` instead of C-style array
- Do not use pointer arithmetic
- Do not use reinterpret_cast
- etc.

## How to check this PR
```shell
# build the package
# edit hdd.service
sudo systemctl restart hdd.service
# tty 1
ros2 launch system_monitor system_monitor.launch.xml
# tty 2
ros2 run rqt_runtime_monitor rqt_runtime_monitor
```
Seems to work correctly in my environment.
Screenshot from 2021-04-02 17-42-54
Screenshot from 2021-04-02 17-43-11

### DIFF
--- a/system/system_monitor/reader/hdd_reader/hdd_reader.cpp
+++ b/system/system_monitor/reader/hdd_reader/hdd_reader.cpp
@@ -17,14 +17,14 @@
  * @brief HDD information read class
  */
 
-#include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
 #include <syslog.h>
 #include <unistd.h>
+#include <cerrno>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
 
 #include <algorithm>
 #include <regex>
@@ -47,20 +47,18 @@
 
 #include "hdd_reader/hdd_reader.hpp"
 
-namespace fs = boost::filesystem;
-
 // 7634-7647 Unassigned
 constexpr int PORT = 7635;
 
 /**
  * @brief HDD information
  */
-typedef struct
+using HDD_Info = struct
 {
   std::string model_;   //!< @brief Model number
   std::string serial_;  //!< @brief Serial number
   int temperature_;     //!< @brief Temperature
-} HDD_Info;
+};
 
 /**
  * @brief ATA PASS-THROUGH (12) command
@@ -68,7 +66,7 @@ typedef struct
  * - ATA Command Pass-Through
  *   https://www.t10.org/ftp/t10/document.04/04-262r8.pdf
  */
-typedef struct
+using ATAPassThrough12 = struct
 {
   uint8_t operation_code_;      //!< @brief OPERATION CODE (A1h)
   uint8_t reserved0_ : 1;       //!< @brief Reserved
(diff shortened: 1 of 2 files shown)


system/system_monitor/reader/hdd_reader/hdd_reader.cpp
```diff
@@ -47,28 +47,26 @@
...
-namespace fs = boost::filesystem;
-
 // 7634-7647 Unassigned
 constexpr int PORT = 7635;
 
 /**
  * @brief HDD information
  */
-typedef struct
+using HDD_Info = struct
```
kenji-miyake, line 56: or just `struct HDD_Info {};` ?

system/system_monitor/reader/hdd_reader/hdd_reader.cpp
```diff
@@ -89,15 +87,15 @@ typedef struct
...
+};
 
 /**
  * @brief Attribute Table Format
  * @note For details please see the documents below.
  * - SMART Attribute Overview
  *   http://www.t13.org/Documents/UploadedDocuments/docs2005/e05171r0-ACS-SMARTAttributes_Overview.pdf
  */
-typedef struct __attribute__((packed))  // Minimize total struct memory 16 to 12
+using AttributeEntry = struct __attribute__((packed))  // Minimize total struct memory 16 to 12
```
kenji-miyake, line 98: How about using `#pragma pack`? Is it different from `__attribute__((packed))`?

ito-san: The usage of `#pragma pack(1)` is better style, I think.

kmiya: The `__attribute__((packed))` might be a GCC-specific in ancient times.
Nowadays, we can use it in Clang as well as GCC. Seems to be no problem.
https://wandbox.org/permlink/p1fWzK5JqEVVDnaJ
Also, AUTOSAR C++14 says `#pragma` should not be used (Rule A16-7-1).

system/system_monitor/reader/hdd_reader/hdd_reader.cpp
kmiya: Oh, yeah! Fixed in 9a9612f83ad86dc2f070485d189f567a317b0f97.
`packed` still works.
https://wandbox.org/permlink/hZxGP08FUcCF1BLP

system/system_monitor/reader/hdd_reader/hdd_reader.cpp
ito-san: Understood. Looks like it should stay as it is! :+1:

kenji-miyake (approved):

PR #731, opened 2021-11-25
Fix unstable test in trajectory_follower_nodes

KeisukeShima:
## Description

I was trying to figure out why gtests were sometimes failing on CI.
As a result, I determined that I need to fix the testing method.

The `trajectory_follower` assumes that there are enough messages in the `tf_buffer` before the trajectory message is received.
In our tests, we did not process messages using `spin_some()` after `sendTransform()`. This caused a problem where the order of arrival of the messages was swapped, especially under heavy CPU load, and the current position in `trajectory_follower` could not be obtained.
To fix this, I added a `spin_some()` and debug messages.

## Review Procedure

To check this issue on local environment, follow this step:

Terminal1:
`stress-ng -c 8 -l 90`
This simulates 90% CPU load, each core. (note that my CPU has 8 cores. Edit this option if you need.)

Terminal2:
`colcon test --packages-select trajectory_follower_nodes --event-handlers console_cohesion+`

## Pre-Review Checklist for the PR Author

**PR Author should check the checkboxes below when creating the PR.**

- [x] Read [pull request guidelines][pull-request-guidelines]
- [x] Code follows [coding guidelines][coding-guidelines]
- [x] Assign PR to reviewer

## Checklist for the PR Reviewer

**Reviewers should check the checkboxes below before approval.**

- [ ] Commits are properly organized and messages are according to the guideline
- [ ] Code follows [coding guidelines][coding-guidelines]
- [ ] (Optional) Unit tests have been written for new behavior
- [ ] PR title describes the changes

## Post-Review Checklist for the PR Author

**PR Author should check the checkboxes below before merging.**

- [x] All open points are addressed and tracked via issues or tickets
- [ ] Write [release notes][release-notes]

## CI Checks

- **Build and test for PR / build-and-test-pr**: Required to pass before the merge.
- **Build and test for PR / clang-tidy-pr**: NOT required to pass before the merge. It is up to the reviewer(s). Found false positives? See the [guidelines][clang-tidy-guidelines].
- **Check spelling**: NOT required to pass before the merge. It is up to the reviewer(s). See [here][spell-check-dict] if you want to add some words to the spell check dictionary.

[clang-tidy-guidelines]: https://tier4.github.io/autoware.proj/tree/main/developer_guide/ClangTidyGuideline/
[coding-guidelines]: https://tier4.atlassian.net/wiki/spaces/AIP/pages/1194394777/T4
[pull-request-guidelines]: https://tier4.github.io/autoware.proj/tree/main/developer_guide/PullRequestGuideline/
[release-notes]: https://tier4.atlassian.net/wiki/spaces/AIP/pages/563774416
[spell-check-dict]: https://github.com/tier4/autoware-spell-check-dict#how-to-contribute

### DIFF
--- a/control/trajectory_follower_nodes/src/lateral_controller_node.cpp
+++ b/control/trajectory_follower_nodes/src/lateral_controller_node.cpp
@@ -265,7 +265,7 @@ void LateralController::onTrajectory(
   m_current_trajectory_ptr = msg;
 
   if (!m_current_pose_ptr && !updateCurrentPose()) {
-    RCLCPP_DEBUG(get_logger(), "Current pose is invalid!!");
+    RCLCPP_DEBUG(get_logger(), "Current pose is not received yet.");
     return;
   }
 
--- a/control/trajectory_follower_nodes/src/lateral_controller_node.cpp
+++ b/control/trajectory_follower_nodes/src/lateral_controller_node.cpp
@@ -181,7 +181,6 @@ LateralController::~LateralController()
 void LateralController::onTimer()
 {
   if (!checkData() || !updateCurrentPose()) {
-    publishCtrlCmd(getStopControlCommand());
     return;
   }
 
@@ -266,6 +265,7 @@ void LateralController::onTrajectory(
   m_current_trajectory_ptr = msg;
 
   if (!m_current_pose_ptr && !updateCurrentPose()) {
+    RCLCPP_DEBUG(get_logger(), "Current pose is invalid!!");
     return;
   }
 
--- a/control/trajectory_follower_nodes/test/test_lateral_controller_node.cpp
+++ b/control/trajectory_follower_nodes/test/test_lateral_controller_node.cpp
@@ -89,10 +89,8 @@ TEST_F(FakeNodeFixture, no_input)
     std::make_shared<tf2_ros::StaticTransformBroadcaster>(this->get_fake_node());
 
   // No published data: expect a stopped command
-  test_utils::waitForMessage(node, this, received_lateral_command);
-  ASSERT_TRUE(received_lateral_command);
-  EXPECT_EQ(cmd_msg->steering_tire_angle, 0.0f);
-  EXPECT_EQ(cmd_msg->steering_tire_rotation_rate, 0.0f);
+  test_utils::waitForMessage(node, this, received_lateral_command, std::chrono::seconds{1LL}, false);
+  ASSERT_FALSE(received_lateral_command);
 }
 
 TEST_F(FakeNodeFixture, empty_trajectory)
@@ -140,11 +138,8 @@ TEST_F(FakeNodeFixture, empty_trajectory)
   odom_pub->publish(odom_msg);
   steer_pub->publish(steer_msg);
 
-  test_utils::waitForMessage(node, this, received_lateral_command);
-  ASSERT_TRUE(received_lateral_command);
-  EXPECT_EQ(cmd_msg->steering_tire_angle, 0.0f);
-  EXPECT_EQ(cmd_msg->steering_tire_rotation_rate, 0.0f);
-  EXPECT_GT(rclcpp::Time(cmd_msg->stamp), rclcpp::Time(traj_msg.header.stamp));
+  test_utils::waitForMessage(node, this, received_lateral_command, std::chrono::seconds{1LL}, false);
+  ASSERT_FALSE(received_lateral_command);
 }
 
 TEST_F(FakeNodeFixture, straight_trajectory)
@@ -177,6 +172,10 @@ TEST_F(FakeNodeFixture, straight_trajectory)
   geometry_msgs::msg::TransformStamped transform = test_utils::getDummyTransform();
   transform.header.stamp = node->now();
   br->sendTransform(transform);
+
+  // Spin for transform to be published
+  test_utils::spinWhile(node);
+
   // Straight trajectory: expect no steering
   received_lateral_command = false;
   Trajectory traj_msg;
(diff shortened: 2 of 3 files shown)


control/trajectory_follower_nodes/test/test_lateral_controller_node.cpp
```diff
@@ -280,11 +287,11 @@ TEST_F(FakeNodeFixture, right_turn)
   odom_pub->publish(odom_msg);
   steer_pub->publish(steer_msg);
 
-  test_utils::spinWhile(node);
   test_utils::waitForMessage(node, this, received_lateral_command);
   ASSERT_TRUE(received_lateral_command);
   EXPECT_LT(cmd_msg->steering_tire_angle, 0.0f);
-  EXPECT_LT(cmd_msg->steering_tire_rotation_rate, 0.0f);
+  // EXPECT_LT(cmd_msg->steering_tire_rotation_rate, 0.0f);  // NOTE: Is the statement correct?
+  // 1: Expected: (cmd_msg->steering_tire_rotation_rate) < (0.0f), actual: 5.55112e-16 vs 0
```
KeisukeShima, line 294: Question: Is this statement correct? I think steering tire rotation rate has positive value.

maxime-clem: From my understanding the steering is supposed to be negative in order to turn to the right.
In the following test the steering is positive for a left turn.

control/trajectory_follower_nodes/src/lateral_controller_node.cpp
```diff
@@ -181,7 +181,6 @@ LateralController::~LateralController()
 void LateralController::onTimer()
 {
   if (!checkData() || !updateCurrentPose()) {
-    publishCtrlCmd(getStopControlCommand());
```
rej55, line 184: @TakaHoribe Is it ok not to publish stop control command when it doesn't receive the data?
If we apply this change, it doesn't publish any control command when this situation.

control/trajectory_follower_nodes/test/test_lateral_controller_node.cpp
KeisukeShima: @maxime-clem Thanks for your advice! I will investigate why this test fails.

control/trajectory_follower_nodes/src/lateral_controller_node.cpp
KeisukeShima: I made the change based on a discussion in Slack. If this change is not needed, it can be reverted.

kenji-miyake: Publishing stop control command in this layer isn't good because it cannot be taken into account in the fail-safe system.
So I believe it should stop publishing commands and let the fail-safe system know the error.

rej55: OK. I agree your comment. Thank you!
(I just worried about changing behavior with this change.)

control/trajectory_follower_nodes/test/test_lateral_controller_node.cpp
KeisukeShima: When I tried again, the test worked fine. Perhaps my local environment was broken.

taikitanaka3: it looks like only control command that is calculated in stopControlCmd() is used so i think  cmd_msg->steering_tire_rotation_rate becomes almost 0.0
https://github.com/tier4/AutowareArchitectureProposal.iv/blob/0a47ea8475b7a0cd0ed8648dcc2ce0011e21992d/control/trajectory_follower_nodes/src/lateral_controller_node.cpp#L177

KeisukeShima: @taikitanaka3 Thanks for your comment!
In `right_turn` test, `ctrl_cmd` is published inside `onTimer()` function.

https://github.com/tier4/AutowareArchitectureProposal.iv/blob/0a47ea8475b7a0cd0ed8648dcc2ce0011e21992d/control/trajectory_follower_nodes/src/lateral_controller_node.cpp#L223

I think the destructor will not be called while the test is running because the `trajectory_follower_node` is not terminated.

control/trajectory_follower_nodes/src/lateral_controller_node.cpp
```diff
@@ -266,6 +265,7 @@ void LateralController::onTrajectory(
   m_current_trajectory_ptr = msg;
 
   if (!m_current_pose_ptr && !updateCurrentPose()) {
+    RCLCPP_DEBUG(get_logger(), "Current pose is invalid!!");
```
TakaHoribe, line 268: It is not invalid, just not received yet?

```suggestion
    RCLCPP_DEBUG(get_logger(), "Current pose is not received yet.");
```

KeisukeShima: Thanks. It's better.

control/trajectory_follower_nodes/test/test_lateral_controller_node.cpp
taikitanaka3: @KeisukeShima 
Ah yes your right.

control/trajectory_follower_nodes/src/lateral_controller_node.cpp
TakaHoribe: @rej55 @kenji-miyake Could you write here how the system behavior will change by this fix in current products? Just for in case.
Since JpnTaxi doesn't see some diags, it would be a bad situation.

taikitanaka3 (approved):
LGTM

TakaHoribe:
影響範囲確認したいので、一旦マージ待ってください。https://github.com/tier4/AutowareArchitectureProposal.iv/pull/731#discussion_r756630721

control/trajectory_follower_nodes/src/lateral_controller_node.cpp
TakaHoribe: The stop command from MPC is overwritten by the velocity control, so it has no effect. Excuse me.

TakaHoribe (approved):

